Issue #343, opened 2014-02-25
Download hearing impaired subtitles optionally


h3llrais3r:
Is is possible to change the logic for the hearing impaired?
Now the hearing impaired flag of the subtitle must match the hearing impaired flag specified in the download_best_subtitles(...)

<code>
if subtitle.hearing_impaired != hearing_impaired:
&nbsp;&nbsp;logger.debug('Skipping subtitle: hearing impaired != %r', hearing_impaired)
&nbsp;&nbsp;continue
</code>

Could this be adapted as mentioned hereunder:
hearing_impaired == False -> download only subtitles with hearing_impaired == False
hearing_impaired == True -> download subtitles with both hearing_impaired == False AND hearing_impaired == True

I you want to keep the possibility to dowload only hearing_impaired subtitles, it could also be adapted to an int (0, 1, 2) instead of a boolean (False and True).

olfway:
Any news on this?

h3llrais3r:
Waiting for merging of PR https://github.com/Diaoul/subliminal/pull/361...

Issue #268, opened 2013-11-14
Subliminal 0.8


Diaoul:
API Changes
-------------------
- [x] Add `content` and encoding related stuff to the `Subtitle` class
- [x] Add a `save_subtitles` function that will save the subtitles on the filesystem based on the `Video` name, the `single` parameter and the `directory` parameter
- [x] Add a `page_link` attribute to the `Subtitle` class that links to the page where the subtitle can be downloaded manually (if applicable, `None` otherwise)
- [x] Add a `ProviderManager` that handle multiple providers, filtering (by name, video_types and languages), initialization, termination and retrying
- [x] Add a `year` to `Episode`, make sure no year is a match
- [x] Refactor current API with the above changes
- [x] Refactor Exceptions
- [x] Add timeout to ServerProxy #254
- [x] Add a `format` to `Video` and use it for matching
- [x] Use babelfish 0.5.0
- [x] Make the equivalent of ConverterManager in babelfish for Provider

Diaoul:
This year. :laughing:

h3llrais3r:
hehe :

Diaoul:
Joke aside, I'm doing a lot of bug fixing and unit testing. There won't be any 0.8.0 version, jumping directly to 1.0 as I think the API is stabilized and the code battletested.

About Hearing Impaired, I think the proper place is in `compute_matches`, the priority is to have best matching subtitles so HI will be the lowest on the scale.

I'll update the code accordingly.

h3llrais3r:
Nice te hear that a stable 1.0 version will be released! ;)

About the hearing impaired: are you talking about https://github.com/Diaoul/subliminal/pull/361/files?
If yes, will you be adding a score for HI?
With my pull request, the user can decide himself wether he wants HI subtitles or not or does not care about it.
-> perhaps it's better to discuss in the PR?

PR #361, opened 2014-03-24
Make hearing_impaired option more flexible

h3llrais3r:
Hearing impaired has now 3 values:
- True -> download only hearing_impaired subtitles
- False -> download no hearing_impaired subtitles
- None -> download all (do not check hearing_impaired value)

-> None becomes default value in download_best_subtitles

-> Fixes https://github.com/Diaoul/subliminal/issues/343

### DIFF
--- a/subliminal/api.py
+++ b/subliminal/api.py
@@ -54,7 +54,7 @@ def download_subtitles(subtitles, provider_configs=None):
 
 
 def download_best_subtitles(videos, languages, providers=None, provider_configs=None, min_score=0,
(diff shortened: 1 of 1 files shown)


Diaoul:
There are some things about hearing impared merged in 0.7.x I believe. I need to backport a few fixes to 0.8.0 and merge this afterward.

Diaoul:
Following our discussion in #268, I want to make this behavior optional too. I'll see what I can do with the `compute_match` function.

PR #381, opened 2014-05-20
Fix brazilian portuguese content decode

leobragatti:


Diaoul:
Can you open an issue with reproduce steps? I'll add some unit tests

Issue #378, opened 2014-05-08
subliminal does not find any subtitles for "Marvel's Agents of S.H.I.E.L.D" on Addic7ed


SeeDborg:
Everything is in the title.
Every time I try, this is what I get : 

INFO     [subliminal.api] Listing subtitles with provider 'addic7ed' for video &lt;Episode [u"Marvel's Agents of S H I E L D", 1x21]> with languages set([&lt;Language [fr]>])
INFO     [subliminal.api] Found 0 subtitles
INFO     [subliminal.api] Listing subtitles with provider 'addic7ed' for video &lt;Episode [u"Marvel's Agents of S H I E L D", 1x20]> with languages set([&lt;Language [fr]>])
INFO     [subliminal.api] Found 0 subtitles
INFO     [subliminal.providers.addic7ed] Logged out
No subtitles downloaded

Issue #334, opened 2014-02-11
SyntaxError: invalid syntax ?


meokey:
$ subliminal -l en -- White\ House\ Down.(2013).mkv
Traceback (most recent call last):
  File "/usr/bin/subliminal", line 9, in <module>
    load_entry_point('subliminal==0.7.4', 'console_scripts', 'subliminal')()
  File "build/bdist.linux-x86_64/egg/pkg_resources.py", line 353, in load_entry_point
    `manager` must be an `IResourceManager`"""
  File "build/bdist.linux-x86_64/egg/pkg_resources.py", line 2302, in load_entry_point

  File "build/bdist.linux-x86_64/egg/pkg_resources.py", line 2029, in load
    if group is None:
  File "/usr/lib/python2.6/site-packages/subliminal/**init**.py", line 9, in <module>
    from .api import PROVIDERS_ENTRY_POINT, list_subtitles, download_subtitles, download_best_subtitles
  File "/usr/lib/python2.6/site-packages/subliminal/api.py", line 95
    providers_by_name = {ep.name: ep.load() for ep in pkg_resources.iter_entry_points(PROVIDERS_ENTRY_POINT)}
^ (this ^ is under letter r in 'for')
SyntaxError: invalid syntax

PS. there is error when installing

$ sudo pip install subliminal
Downloading/unpacking subliminal
  Downloading subliminal-0.7.4.tar.gz
  Running setup.py (path:/tmp/pip_build_root/subliminal/setup.py) egg_info for package subliminal
Downloading/unpacking beautifulsoup4>=4.3.2 (from subliminal)
  Downloading beautifulsoup4-4.3.2.tar.gz (143kB): 143kB downloaded
  Running setup.py (path:/tmp/pip_build_root/beautifulsoup4/setup.py) egg_info for package beautifulsoup4
Downloading/unpacking guessit>=0.6.2,<0.7 (from subliminal)
  Downloading guessit-0.6.2.tar.gz (123kB): 123kB downloaded
  Running setup.py (path:/tmp/pip_build_root/guessit/setup.py) egg_info for package guessit
Requirement already satisfied (use --upgrade to upgrade): requests>=2.0.1 in /usr/lib/python2.6/site-packages (from subliminal)
Downloading/unpacking enzyme>=0.4.0 (from subliminal)
  Downloading enzyme-0.4.1.tar.gz (47kB): 47kB downloaded
[...]

PR #404, opened 2014-09-18
0.7.x

caronc:
I've been [maintaining a plugin](https://github.com/caronc/nzbget-subliminal) ([forum](http://nzbget.net/forum/viewtopic.php?f=8&t=1422&sid=9cc9c3414699137f3058f570caadfe10)) to your fantastic subtitle fetcher you've written. The  [plugin](https://github.com/caronc/nzbget-subliminal) is for [NZBGet](http://nzbget.net) (a [UseNet](http://en.wikipedia.org/wiki/Usenet) solution). You can [see this blog if interested](http://nuxref.com/2014/07/15/a-usenet-solution-centos-6/). I'm only utilizing the 0.7.x branch, but have added some stability to it along with back porting older code so that it works with people running older versions of python (v2.6+) (like myself). The backporting specifically addresses [an old ticket closed](https://github.com/Diaoul/subliminal/issues/334). I adapted to the podnapsi's new page layout and added a few other minor features. I just thought I should share my changes back!

Just added support for quotes which should fix [issue 378](https://github.com/Diaoul/subliminal/issues/378) and [issue 412](https://github.com/Diaoul/subliminal/issues/412) for you.


caronc:
FYI, testing failures have nothing to do with my code changes.
The first 4 are errors with guessit,  the rest are DNS resolution errors (dns issue with [coveralls](https://travis-ci.org/Diaoul/subliminal/builds/48295043) perhaps)?.  I'm presently using guessit v0.9.3 without any problems at all.

subliminal/api.py
```diff
@@ -6,12 +6,34 @@
...
+    'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/41.0.2227.0 Safari/537.36',
+    'Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/41.0.2227.0 Safari/537.36',
+    'Mozilla/5.0 (compatible, MSIE 11, Windows NT 6.3; Trident/7.0; rv:11.0) like Gecko',
+    'Mozilla/5.0 (compatible; MSIE 10.0; Windows NT 6.1; WOW64; Trident/6.0)',
+    'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_9_3) AppleWebKit/537.75.14 (KHTML, like Gecko) Version/7.0.3 Safari/7046A194A',
+    'Mozilla/5.0 (iPad; CPU OS 6_0 like Mac OS X) AppleWebKit/536.26 (KHTML, like Gecko) Version/6.0 Mobile/10A5355d Safari/8536.25',
+)
+
+# Returns a random agent to use from the list above
+RANDOM_USER_AGENT = AGENT_LIST[randint(0, len(AGENT_LIST)-1)]
```
Diaoul: This is set at import time, better randomize at runtime

subliminal/api.py
```diff
@@ -214,30 +275,38 @@ def download_best_subtitles(videos, languages, providers=None, provider_configs=
                                 provider_name, video, provider_video_languages)
                     try:
                         provider_subtitles = provider.list_subtitles(video, provider_video_languages)
-                    except ProviderNotAvailable:
+                    except ProviderNotAvailable as err:
                         logger.warning('Provider %r is not available, discarding it', provider_name)
+                        logger.debug('ProviderNotAvailable error: %r', str(err))
```
Diaoul: You can use `exc_info=True` as kwargs for logger.debug to add trace information

subliminal/api.py
```diff
@@ -136,8 +183,9 @@ def download_subtitles(subtitles, provider_configs=None, single=False):
                     continue
                 with io.open(subtitle_path, 'w', encoding='utf-8') as f:
                     f.write(subtitle_text)
-                downloaded_languages.add(subtitle.language)
-                if single or downloaded_languages == languages:
+                    downloaded_languages.add(subtitle.language)
+                    fetched_subtitles.add(basename(subtitle_path))
```
Diaoul: The language check should be enough here, why the fetched_subtitles check?

subliminal/cli.py
```diff
@@ -52,11 +52,6 @@ def subliminal():
     filtering_group.add_argument('-f', '--force', action='store_true',
                                  help='force subtitle download for videos with existing subtitles')
 
-    # addic7ed
-    addic7ed_group = parser.add_argument_group('addic7ed')
-    addic7ed_group.add_argument('--addic7ed-username', metavar='USERNAME', help='username for addic7ed provider')
-    addic7ed_group.add_argument('--addic7ed-password', metavar='PASSWORD', help='password for addic7ed provider')
-
```
Diaoul: Why remove this?

subliminal/providers/addic7ed.py
```diff
@@ -53,49 +55,21 @@ def compute_matches(self, video):
...
-            except requests.Timeout:
-                raise ProviderNotAvailable('Timeout after 10 seconds')
-            if r.status_code != 200:
-                raise ProviderNotAvailable('Request failed with status code %d' % r.status_code)
-        self.session.close()
+        #self.session.headers = {'User-Agent': 'Subliminal/%s' % __version__}
+        self.session.headers = {
+            'User-Agent': RANDOM_USER_AGENT,
+            'Referer': self.server,
+        }
```
Diaoul: Why use a random user agent here? I'm generally against this because servers should be allowed to block subliminal activity. Although I don't want them too, it should be possible. We should contact the provider to know the reason why they would want to block us and have an agreement.

subliminal/providers/addic7ed.py
```diff
@@ -108,7 +82,7 @@ def get(self, url, params=None):
 
         """
         try:
-            r = self.session.get(self.server + url, params=params, timeout=10)
+            r = self.session.get(self.server + url, params=params, timeout=30)
```
Diaoul: I think 10s timeout is plenty, if the provider doesn't respond by then, they are experiencing serious trouble.

subliminal/providers/addic7ed.py
```diff
@@ -126,7 +100,9 @@ def get_show_ids(self):
         soup = self.get('/shows.php')
         show_ids = {}
         for html_show in soup.select('td.version > h3 > a[href^="/show/"]'):
-            show_ids[html_show.string.lower()] = int(html_show['href'][6:])
+            show_ids[
+                IGNORED_CHARACTERS_RE.sub('', html_show.string).lower()] = \
+                    int(html_show['href'][6:])
```
Diaoul: You should rename `IGNORED_CHARACTERS_RE` to something more cache-specific and create a `sanitize_cache` function (or similar) to put this logic.

subliminal/providers/addic7ed.py
```diff
@@ -150,10 +126,11 @@ def find_show_id(self, series):
 
     def query(self, series, season):
         show_ids = self.get_show_ids()
-        if series.lower() in show_ids:
-            show_id = show_ids[series.lower()]
+        _series = IGNORED_CHARACTERS_RE.sub('', series).lower()
```
Diaoul: This name is confusing, please make this explicit such as `sanitized_series`

subliminal/providers/addic7ed.py
```diff
@@ -180,7 +157,7 @@ def list_subtitles(self, video, languages):
 
     def download_subtitle(self, subtitle):
         try:
-            r = self.session.get(self.server + subtitle.download_link, timeout=10,
+            r = self.session.get(self.server + subtitle.download_link, timeout=30,
```
Diaoul: Same comment about the timeout here.

subliminal/providers/podnapisi.py
```diff
@@ -41,7 +46,9 @@ def compute_matches(self, video):
         # episode
         if isinstance(video, Episode):
             # series
-            if video.series and self.series.lower() == video.series.lower():
+            if video.series and \
+               IGNORED_CHARACTERS_RE.sub('', self.series).lower() == \
+               IGNORED_CHARACTERS_RE.sub('', video.series).lower():
```
Diaoul: Maybe worth adding a `sanitized_series` property on the `video` object what do you think?

subliminal/providers/podnapisi.py
```diff
@@ -116,20 +137,30 @@ def query(self, language, series=None, season=None, episode=None, title=None, ye
...
-                subtitles.extend([PodnapisiSubtitle(language, int(s.find('id').text), s.find('release').text.split(),
+                try:
+                    subtitles.extend([PodnapisiSubtitle(language, int(s.find('id').text), s.find('release').text.split(),
                                                     'h' in (s.find('flags').text or ''), s.find('url').text[38:],
                                                     series=series, season=season, episode=episode)
                                   for s in root.findall('subtitle')])
+                except AttributeError:
+                    # there simply wasn't enough information in the TV Show
+                    # gracefully handle this instead of crashing :)
+                    break
```
Diaoul: Have you added a unittest for this? I think this is nice to handle it but maybe we need to be more specific about it by outputing a log message.

subliminal/providers/podnapisi.py
```diff
@@ -153,7 +199,7 @@ def download_subtitle(self, subtitle):
             raise ProviderNotAvailable('Timeout after 10 seconds')
         if r.status_code != 200:
             raise ProviderNotAvailable('Request failed with status code %d' % r.status_code)
-        with zipfile.ZipFile(io.BytesIO(r.content)) as zf:
+        with contextlib.closing(zipfile.ZipFile(io.BytesIO(r.content))) as zf:
```
Diaoul: This is why I hate python 2.6...

subliminal/providers/tvsubtitles.py
```diff
@@ -105,15 +108,35 @@ def find_show_id(self, series):
...
+        _series = IGNORE_DATEMATCH.match(
+            IGNORED_CHARACTERS_RE.sub('', series)\
+                .replace('.', ' ').strip().lower(),
+        )
+        if not _series:
+            _series = IGNORED_CHARACTERS_RE.sub('', series)\
+                        .replace('.', ' ').strip().lower()
+        else:
+            _series = _series.group(1)
+
```
Diaoul: A little comment on this and an explicit name instead of `_series` would be nice :star2:

subliminal/providers/tvsubtitles.py
```diff
@@ -105,15 +108,35 @@ def find_show_id(self, series):
...
+            show = IGNORE_DATEMATCH.match(
+                IGNORED_CHARACTERS_RE.sub('', match.group('series'))\
+                        .replace('.', ' ').strip().lower(),
+            )
+            if not show:
+                logger.warning('Could not postparse %r', match.group('series'))
+                continue
+            show = show.group(1)
+
+            if show == _series:
```
Diaoul: Same here

subliminal/tests/test_providers.py
```diff
@@ -30,131 +30,61 @@ def test_find_show_id(self):
     def test_find_show_id_error(self):
         with self.Provider() as provider:
             show_id = provider.find_show_id('the big how i met your mother')
-        self.assertIsNone(show_id)
+        self.assertEqual(show_id, None)
```
Diaoul: 2.6 compat strikes again...

subliminal/video.py
```diff
@@ -147,6 +163,18 @@ def __repr__(self):
             return '<%s [%r]>' % (self.__class__.__name__, self.title)
         return '<%s [%r, %r]>' % (self.__class__.__name__, self.title, self.year)
 
+    def __hash__(self):
+        if self.year is None:
+            return hash((
+                self.title,
+                self.year,
+            ))
+        return hash(self.title)
```
Diaoul: Is this used anywhere?

subliminal/video.py
```diff
@@ -147,6 +163,18 @@ def __repr__(self):
...
+
+    def __eq__(self, other):
+        return self.__class__.__name__ == other.__class__.__name__\
+                and self.title == other.title\
+                and self.year == other.year
```
Diaoul: Or this?

Diaoul:
I just review your changes. There are just a few points:
* Python 2.6 is from eons ago, it makes the code less readable and not exactly python3 friendly, would you agree to drop this? What system are you running subliminal on to be that outdated?
* I'd like to merge and publish this as a new (and hopefully last) release of the 0.7.x branch and rebase 0.8 on this.

caronc:
I realize Python v2.6 is old, but I'm old fashioned too. I'd trade stability over bleeding edge any day. I run Red Hat 6.6 on all my production systems at work and CentOS 6.6 at home. As a result, I'm stuck with it's shipped version of Python 2.6.  I've even packaged all of your software in [RPMs](http://repo.lead2gold.org/centos/6/en/x86_64/custom/repoview/python-subliminal.html) which makes it so easy to deploy your fantastic tool. :)

I intend on updating to CentOS 7 soon which will bring me to Python v2.7.  I'll still be old; but we'll at least be on the same page.  :)

I'll read your review notes, I'll reply to them individually and do my best to explain them; from there you can let me know how you want me to proceed.

subliminal/video.py
caronc: At one point i was using it for the [NZBGet plugin](nzbget.net/forum/viewtopic.php?f=8&t=1422) I wrote. It allows you to **filter()** content and eliminate duplicate videos found.  the **==** operator or **in** operator could be used when comparing a big list of videos.

I can remove it if you want; i realize your not using it internally for anything It just makes it easier for people (like me) who use your tool's lower level functions to handle batches etc.

subliminal/providers/podnapisi.py
caronc: Let me see what i can do about this.  There should be no reason i can't accommodate both of your very valid points.

subliminal/providers/addic7ed.py
caronc: Makes sense; i totally agree with the naming convention your suggestion.

subliminal/providers/addic7ed.py
caronc: I think i had a failure one day and got lazy and changed the value;  i'll flip it back.

subliminal/providers/addic7ed.py
caronc: I totally agree with you but Addic7ted has recently blocked all access their site 'unless' we do this.  I even forked my nzbget-subliminal code into using the name **nokarma** and [explain here why](http://nzbget.net/forum/viewtopic.php?f=8&t=1422&start=190#p11683).  It is part of my later commits to this branch.  

In short, I've attempted to make contact with them and after a week (now it's been about 3) without hearing back from them; i just did this dirty move so we could still use the provider. I can roll this back if you want; it will break this provider though (see #428).

subliminal/cli.py
caronc: see my comment on your next review about this provider (with respect to #428)

subliminal/api.py
caronc: This way it doesn't change for the entire instance your using it.  I wouldn't want 1 request to come from what appears to be 1 browser, and the second from another.  A random one is chosen at the start and remains so until the actions you issued are complete.  It was kind of my intentions. Did you want me to change this then to be at runtime instead?

Diaoul: It's OK to have it randomized but import time is too early IMO. If a long running application imports subliminal the User-Agent would be fixed once and for all, maybe add a user-agent parameter to the Addic7ed `__init__` and use a random one if none is provided? That way it's still leaves the ability for end users to customize.

subliminal/video.py
Diaoul: I'm OK to keep it as it makes total sense for library users, I didn't see direct use of this in the library so I just wondered why add this :)

Diaoul:
I don't mind the dict and set comprehension stuff for 2.6 compat, what bothers me is more unittests that are less readable (and outputs less readable failures) and the zipfile stuff.

Maybe for 2.6 compat we can use 2.7+ syntax with [unittest2](https://pypi.python.org/pypi/unittest2)?

caronc:
@sioban: just to avoid confusion, an option as you suggest would prove to be unnecessary (IMO).  At this time, the proper User-Agent (*Subliminal v0.7*) is correctly applied to all providers except Addic7ed (as of very recently).  If I were to revert back and apply the same User-Agent for Addic7ed too (as an option), then the provider simply won't work.
@Diaoul: I've accommodated most of the core requests you made, there are still some lingering I'll try to get to. Travis CI is really hit and miss eh? Today i passed, but yesterday i fail because providers are unavailable from the testing host. Oh well; right now things are working great, so that's a good thing! :)

Diaoul:
That's the problem with the unittests. I think the best for unittest would be to have a mock proxy that forwards the request to the original website, caches the response and send it back to the client *as-is*. If the original website fails to respond, the cache is used. So instead of testing against *real* servers we test against *real or last time* servers. That should be enough.

I can host this mock proxy locally or on *free* Heroku dyno I think.

I'll merge this and eventually make a few changes so I can make a new bugfix release of subliminal this week.

caronc:
Sure! Thanks for accepting the pull request btw. I'm looking forward to dropping all my patch files! :)

j-manu:
@Diaoul What you describe - a mock proxy - is handled in ruby by a gem called VCR. There are two python ports for it - https://github.com/sigmavirus24/betamax & https://github.com/kevin1024/vcrpy

Issue #389, opened 2014-06-25
"Skipping video: Insufficient data to process the guess"


krank:
Updated to latest version in git repo; gives this error on every search:

"Skipping video: Insufficient data to process the guess"

subliminal -l en -v --debug Continuum.S03E13.720p.HDTV.x264-2HD.mkv gives this log:

INFO     [subliminal.video-scan_video:202] Scanning video u'Continuum.S03E13.720p.HDTV.x264-2HD.mkv' in u'/var/downloads/rTorrent/finished/tvshows'
DEBUG    [stevedore.extension-_load_plugins:156] found extension EntryPoint.parse('guess_idnumber = guessit.transfo.guess_idnumber:GuessIdnumber')
DEBUG    [stevedore.extension-_load_plugins:156] found extension EntryPoint.parse('guess_bonus_features = guessit.transfo.guess_bonus_features:GuessBonusFeatures')
DEBUG    [stevedore.extension-_load_plugins:156] found extension EntryPoint.parse('guess_date = guessit.transfo.guess_date:GuessDate')
DEBUG    [stevedore.extension-_load_plugins:156] found extension EntryPoint.parse('split_path_components = guessit.transfo.split_path_components:SplitPathComponents')
DEBUG    [stevedore.extension-_load_plugins:156] found extension EntryPoint.parse('guess_properties = guessit.transfo.guess_properties:GuessProperties')
DEBUG    [stevedore.extension-_load_plugins:156] found extension EntryPoint.parse('guess_release_group = guessit.transfo.guess_release_group:GuessReleaseGroup')
DEBUG    [stevedore.extension-_load_plugins:156] found extension EntryPoint.parse('split_explicit_groups = guessit.transfo.split_explicit_groups:SplitExplicitGroups')
DEBUG    [stevedore.extension-_load_plugins:156] found extension EntryPoint.parse('guess_episodes_rexps = guessit.transfo.guess_episodes_rexps:GuessEpisodesRexps')
DEBUG    [stevedore.extension-_load_plugins:156] found extension EntryPoint.parse('guess_weak_episodes_rexps = guessit.transfo.guess_weak_episodes_rexps:GuessWeakEpisodesRexps')
DEBUG    [stevedore.extension-_load_plugins:156] found extension EntryPoint.parse('guess_episode_info_from_position = guessit.transfo.guess_episode_info_from_position:GuessEpisodeInfoFromPosition')
[...]

PR #405, opened 2014-09-25
Upgrade to guessit 0.9.3

h3llrais3r:
This fixes https://github.com/Diaoul/subliminal/issues/389

The new version of guessit does not automatically check by default for camelCase properties.
If you should need it, it can be added to the guessit options.
See https://github.com/wackou/guessit/issues/160 for more info.


h3llrais3r:
I forgot to update the dependencies and the way to access the guess_properties...

h3llrais3r:
@Diaoul Can you also merge this upgrade?
This fixes some minor issues in guessit.
With my last commits (I forgot some...) the upgrade should be complete. ;)

PR #417, opened 2014-10-29
Add Itasa provider

ningod:
Add Itasa provider and credential configuration from config file


Diaoul:
Please rebase against master and add unit tests.

PR #421, opened 2014-11-22
Fixed unrecognized shows

camon:
There was a problem in get_show_ids that caused show names like "Greys
Anatomy", "NCIS Los Angeles", ... not to be found in show_ids, because
the real names take additional special characters like ' ? : &, so what
i've done was to add extra names to show_ids where basically we remove
or replace the special charecters


Diaoul:
Can you reproduce with master? If so please give reproduce steps. I'll add unit tests.

PR #426, opened 2014-12-27
Also guess video codec from version in addic7ed provider

oxan:
To increase accuracy and for consistency with other providers.


h3llrais3r:
Nice. Apparently I forgot the video codec guess last year... :)

@Diaoul Nice to see you back on the project :+1:

PR #444, opened 2015-03-26
Don't pass empty movie release name to guessit

blissland:
Sometimes the movie_release_name is blank. If so then guessit.guess_movie_info() will crash.


Diaoul:
Can you reproduce with master? If so please tell steps to reproduce in an issue, I'll add unit tests.

PR #463, opened 2015-07-16
Fix for Python 2

martinp:
Python 2 doesn't have the exist_ok parameter.

### DIFF
--- a/subliminal/cli.py
+++ b/subliminal/cli.py
@@ -97,7 +97,11 @@ subliminal_cache = 'subliminal.dbm'
 def subliminal(ctx, addic7ed, cache_dir, debug):
     """Subtitles, faster than your thoughts."""
     # create cache directory
-    os.makedirs(cache_dir, exist_ok=True)
+    try:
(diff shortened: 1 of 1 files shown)


martinp:
Fixed possible race condition.

subliminal/cli.py
```diff
@@ -97,7 +97,11 @@ def convert(self, value, param, ctx):
 def subliminal(ctx, addic7ed, cache_dir, debug):
     """Subtitles, faster than your thoughts."""
     # create cache directory
-    os.makedirs(cache_dir, exist_ok=True)
+    try:
+        os.makedirs(cache_dir)
+    except OSError:
+        if not os.path.isdir(cache_dir):
```
Diaoul: I believe this part is not necessary as it's already done by click. You can just pass.

martinp: I think it's better to raise for any other exceptions that might occur, as OSError is raised for various other errors, e.g. permission errors.

Diaoul: OK, you are right.

PR #485, opened 2015-08-07
Update setup.py

oliver-szymanski:
Upgrade guessit version to get rid of the "IndexError: string index out of range" that prevents found subtitles to be saved.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,7 @@ class PyTest(TestCommand):
         sys.exit(errno)
 
 # requirements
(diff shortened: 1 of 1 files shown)


Diaoul:
Requirements are just here to guarantee API level compatibility. I cannot upgrade subliminal every time a dependency get a fix so I decided on that approach.

Unless you can provide a fairly simple failing unittest with guessit < 0.10.3 I cannot merge this.

PR #489, opened 2015-08-18
Add NapiProjekt provider

### LINKED ISSUE #384: Add provider Napiprojekt
Hello

In Poland none of already available subliminal providers do a good job, even open subtitles can't fine correct subtitles.

We have here provied called NapiProjekt. 

Here is available Synology package which uses pynapi (python).
http://subtitlessh.no-ip.biz Unfortunately it uses MediaInfo lib, which is not available for Intel Atom which is in DS214play.

There are also multiple bash scripts to do that.

I would love to help with integration, as some of docs are only in polish language.

Please let me know if its possible.
Regards
Damian

bogdal:
This PR adds [NapiProjekt](http://www.napiprojekt.pl/) provider.

Fixes #384

### DIFF
--- a/docs/user/cli.rst
+++ b/docs/user/cli.rst
@@ -46,7 +46,7 @@ subliminal download
     Options:
       -l, --language LANGUAGE         Language as IETF code, e.g. en, pt-BR (can be used multiple
                                       times).  [required]
-      -p, --provider [addic7ed|opensubtitles|podnapisi|thesubdb|tvsubtitles]
+      -p, --provider [addic7ed|napiprojekt|opensubtitles|podnapisi|thesubdb|tvsubtitles]
                                       Provider to use (can be used multiple times).
       -a, --age AGE                   Filter videos newer than AGE, e.g. 12h, 1w2d.
       -d, --directory DIR             Directory where to save subtitles, default is next to the video
--- a/docs/user/how_it_works.rst
+++ b/docs/user/how_it_works.rst
@@ -7,6 +7,7 @@ Subliminal uses multiple providers to give users a vast choice and have a better
 subtitles. Current supported providers are:
 
 * Addic7ed
+* NapiProjekt
 * OpenSubtitles
 * Podnapisi
 * TheSubDB
--- a/setup.py
+++ b/setup.py
@@ -80,6 +80,7 @@ setup(name='subliminal',
       entry_points={
           'subliminal.providers': [
               'addic7ed = subliminal.providers.addic7ed:Addic7edProvider',
+              'napiprojekt = subliminal.providers.napiprojekt:NapiProjektProvider',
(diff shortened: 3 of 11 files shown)


Diaoul:
This looks like a very complete PR. I'll review soon, thanks for your work.

singles:
@bogdal Thank you for that PR! Wanted to do this myself - good that I checked PR section first;)

@Diaoul I see that it's merged to master, but I don't see those changes in latest 1.0.1 release - was this removed after all?

Diaoul:
It was merged into master by error. I merged it to develop instead.

castortray:
when we can expect that napiprojekt.py provider will be added to master release ?

Diaoul:
@bogdal: the provider doesn't seem to work anymore. Unittests are failing: https://travis-ci.org/Diaoul/subliminal/jobs/88170450#L395

bogdal:
@Diaoul ok, I will check.

bogdal:
@MichalE7 It looks like `napiprojekt` gives subtitles in a different format than [srt](http://napiprojekt.pl/unit_napisy/dl.php?pass=&t=0ce4a&v=dreambox&l=PL&kolejka=false&napios=Linux&nick=&f=6303e7ee6a835e9fcede9fb2fb00cb36). I assume that we should convert them on the provider level.

MichalE7:
@bogdal on provider level it means where and who can do it? Sickrage, subliminal or napiprojekt itself?
Here is the screenshot of napiprojekt standalone application settings.

napiprojekt_opcje

Based on wikipedia :) srt subtitles should look like the ones downloaded from napiprojekt standalone app: https://en.wikipedia.org/wiki/SubRip

Diaoul:
The format you downloaded with subliminal looks like MicroDVD. Not sure why napiprojekt doesn't return consistent subtitle format, this should be handled by subliminal.

MichalE7:
I did some more checking.
Default setting for standalone app is to download subtitles in 'Original' format so I suppose it downloads subtitles in the format that was uploaded to the database without any conversion. 
In settings you can chose to download subtitles in SRT format which are probably converted from the original. I guess conversion is missing in the script.

Additionally while browsing sites I came across dragon666 project called napi which includes a script to detect and convert subtitles format.
It can be found here: https://github.com/dagon666/napi

Diaoul:
Subliminal can detect a valid/invalid subtitle file with https://github.com/Diaoul/subliminal/blob/e0788be5af94a136e2c875ffcef3ce7d88108626/subliminal/api.py#L157
It should be done when using the CLI, I wonder why it didn't happen here.

I plan to develop a subtitle file format converter for any kind of format with a special focus on TTML for features. For now I think the best is to just discard them somehow and check why the result did get past the is_valid test with proper unittests.

PR #492, opened 2015-08-22
Login opensubtitles

h3llrais3r:
Make it possible to specify your own username and password for downloading from opensubtitles.


mohammad-aqajani:
great job!
how can I use it and set my username and password when import subliminal as a library to my code and then using the download_best_subtitle function? how to pass username and password to this function?

PR #497, opened 2015-08-29
Subscenter support

ofir123:
Updated everything according to the new version.
PR is based from develop instead of master, and the previous one was deleted.


docs/user/cli.rst
```diff
@@ -48,7 +48,7 @@ subliminal download
     Options:
       -l, --language LANGUAGE         Language as IETF code, e.g. en, pt-BR (can be used multiple
                                       times).  [required]
-      -p, --provider [addic7ed|napiprojekt|opensubtitles|podnapisi|thesubdb|tvsubtitles]
+      -p, --provider [addic7ed|napiprojekt|opensubtitles|podnapisi|thesubdb|tvsubtitles|subscenter]
```
Diaoul: The list is sorted. I believe s is before t ;)

setup.py
```diff
@@ -84,7 +84,8 @@ def run_tests(self):
               'opensubtitles = subliminal.providers.opensubtitles:OpenSubtitlesProvider',
               'podnapisi = subliminal.providers.podnapisi:PodnapisiProvider',
               'thesubdb = subliminal.providers.thesubdb:TheSubDBProvider',
-              'tvsubtitles = subliminal.providers.tvsubtitles:TVsubtitlesProvider'
+              'tvsubtitles = subliminal.providers.tvsubtitles:TVsubtitlesProvider',
+              'subscenter = subliminal.providers.subscenter:SubsCenterProvider'
```
Diaoul: Please keep the alphabetical order here

subliminal/cli.py
```diff
@@ -203,6 +203,7 @@ def convert(self, value, param, ctx):
 @click.group(context_settings={'max_content_width': 100}, epilog='Suggestions and bug reports are greatly appreciated: '
              'https://github.com/Diaoul/subliminal/')
 @click.option('--addic7ed', type=click.STRING, nargs=2, metavar='USERNAME PASSWORD', help='Addic7ed configuration.')
+@click.option('--subscenter', type=click.STRING, nargs=2, metavar='USERNAME PASSWORD', help='SubsCenter configuration.')
```
Diaoul: Same here

subliminal/providers/subscenter.py
```diff
@@ -0,0 +1,178 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
```
Diaoul: No more unicode literals

subliminal/providers/subscenter.py
```diff
@@ -0,0 +1,178 @@
...
+import io
+import json
+import logging
+import re
+import zipfile
+import guessit
```
Diaoul: 1 blank line between native and 3rd party imports and each section alphabetically sorted

Diaoul:
Please keep things alphabetically sorted. I'll review the provider soon.

docs/user/cli.rst
ofir123: Done.

setup.py
ofir123: Done.

subliminal/cli.py
ofir123: Done.

subliminal/providers/subscenter.py
ofir123: Done.

subliminal/providers/subscenter.py
ofir123: Done.

ofir123:
Hey!
Fixed some minor things so it would also work on Python 3.
Are you going to merge it soon?
Thnx

Diaoul:
Travis build is not passing, what's the issue?

subliminal/providers/subscenter.py
```diff
@@ -0,0 +1,178 @@
...
+        # title
+        if video.title and self.title and self.title.lower() == video.title.lower():
+            matches.add('title')
+        return matches
+
+
+class SubsCenterProvider(Provider):
+
+    languages = {Language.fromalpha2(l) for l in ['he', 'en']}
+    server = 'http://subscenter.cinemast.com/he/'
```
Diaoul: Why not http://www.subscenter.org/he/?

ofir123: This is the previous URL, and as stated on the main page, it is deprecated.

ofir123:
Fixed all the Travis tests!
Can you please merge?

subliminal/providers/subscenter.py
```diff
@@ -0,0 +1,180 @@
...
+            page_link = self.server + 'subtitle/movie/' + slugified_title + '/'
+        else:
+            raise ValueError('One or more parameters are missing')
+        logger.debug('Searching subtitles %r', {'title': title, 'season': season, 'episode': episode})
+        response = self.session.get(url)
+        if response.status_code != 200:
+            raise ProviderError('Request failed with status code %d' % response.status_code)
+
+        subtitles = []
+        response_json = json.loads(response.content.decode('UTF-8'))
```
ratoaq2: You could use `json.loads(response.text)` and it will use the correct response encoding:

http://docs.python-requests.org/en/latest/user/quickstart/#response-content

subliminal/providers/subscenter.py
```diff
@@ -0,0 +1,180 @@
...
+        elif title:
+            # Search for a movie.
+            kind = 'movie'
+            slugified_title = self.slugify(title)
+            url = self.server + 'cinemast/data/movie/sb/' + slugified_title + '/'
+            page_link = self.server + 'subtitle/movie/' + slugified_title + '/'
+        else:
+            raise ValueError('One or more parameters are missing')
+        logger.debug('Searching subtitles %r', {'title': title, 'season': season, 'episode': episode})
+        response = self.session.get(url)
```
Diaoul: This seems to be a wild guess. I think it's better to use search and cache the result. Is it possible?

subliminal/providers/subscenter.py
```diff
@@ -0,0 +1,180 @@
...
+        title = video.title
+        if isinstance(video, Episode):
+            series = video.series
+            season = video.season
+            episode = video.episode
+        return self.query(languages, series, season, episode, title)
+
+    def download_subtitle(self, subtitle):
+        r = self.session.get(subtitle.download_link, timeout=10, headers={'Referer': subtitle.page_link})
+        if r.status_code != 200:
```
Diaoul: Please use raise_for_status instead.

subliminal/providers/subscenter.py
```diff
@@ -0,0 +1,180 @@
...
+                self.logged_in = True
+            else:
+                raise AuthenticationError(self.username)
+
+    def terminate(self):
+        # logout
+        if self.logged_in:
+            r = self.session.get(self.server + 'subscenter/accounts/logout/', timeout=10)
+            logger.info('Logged out')
+            if r.status_code != 200:
```
Diaoul: Please use raise_for_status instead.

subliminal/providers/subscenter.py
```diff
@@ -0,0 +1,180 @@
...
+            # Search for a movie.
+            kind = 'movie'
+            slugified_title = self.slugify(title)
+            url = self.server + 'cinemast/data/movie/sb/' + slugified_title + '/'
+            page_link = self.server + 'subtitle/movie/' + slugified_title + '/'
+        else:
+            raise ValueError('One or more parameters are missing')
+        logger.debug('Searching subtitles %r', {'title': title, 'season': season, 'episode': episode})
+        response = self.session.get(url)
+        if response.status_code != 200:
```
Diaoul: Please use raise_for_status instead.

subliminal/providers/subscenter.py
```diff
@@ -0,0 +1,180 @@
...
+                for group_data in lang_json.values():
+                    for quality in group_data.values():
+                        for sub in quality.values():
+                            release = sub.get('subtitle_version')
+                            download_link = self.server + 'subtitle/download/' + lang + '/' + str(sub.get('id')) + \
+                                '/?v=' + release + '&key=' + str(sub.get('key'))
+                            subtitles.append(SubsCenterSubtitle(lang_obj, series, season,
+                                                                episode, title, release, kind,
+                                                                bool(sub.get('hearing_impaired', 0)),
+                                                                download_link, page_link))
```
Diaoul: The goal of the SubsCenterSubtitle class is to store all subtitle related stuff as given by the provider. So here you want to store anything that constitutes the download_link without duplicating information. Variables `lang`, `id`, `release` and `key` need to be stored. Not the download link.
You will reconstitute the download link in the `download` method using the information previously stored.

subliminal/providers/subscenter.py
```diff
@@ -0,0 +1,180 @@
...
+        self.session.headers = {'User-Agent': 'Subliminal/%s' % __version__.split('-')[0]}
+        # login
+        if self.username is not None and self.password is not None:
+            logger.debug('Logging in')
+            url = self.server + 'subscenter/accounts/login/'
+            # Retrieve CSRF token first.
+            self.session.get(url)
+            csrf_token = self.session.cookies['csrftoken']
+            data = {'username': self.username, 'password': self.password,
+                    'next': '/he/', 'csrfmiddlewaretoken': csrf_token}
```
Diaoul: Is it necessary to post `next`?

docs/user/how_it_works.rst
```diff
@@ -12,6 +12,7 @@ subtitles. Current supported providers are:
 * Podnapisi
 * TheSubDB
 * TvSubtitles
+* SubsCenter
```
Diaoul: + sorting

Diaoul:
I added a few comments related to the design of the provider.

subliminal/providers/subscenter.py
ofir123: Done!

subliminal/providers/subscenter.py
ofir123: Done!

subliminal/providers/subscenter.py
ofir123: Done!

subliminal/providers/subscenter.py
ofir123: Done!

subliminal/providers/subscenter.py
ofir123: Done!

subliminal/providers/subscenter.py
ofir123: Removed!

docs/user/how_it_works.rst
ofir123: Done!

subliminal/providers/subscenter.py
ofir123: You're only using a cache when querying the server multiple times (for example - when first asking for ID, and then asking for the subtitles themselves).
When working with SubsCenter the query method sends only one request (for the subtitles list), so there's nothing worth caching.

ofir123:
Thank you very much for the extensive CR.
I've fixed everything and the tests are all green :)

subliminal/providers/subscenter.py
Diaoul: The issue here is you guess the URL. It's better to make a search query with the guessed series name and retrieve the URLs from the result page. Is that possible?

ofir123: Ohhh ok I think I understand what you mean.
The problem with this approach is that the site makes you log in before presenting the options, and since I prefer the login part to always be optional, I think this is the best way.

Diaoul: Search doens't require login: http://www.subscenter.org/he/subtitle/search/?q=game+of+thrones
You can iter over the result and retrieve the corresponding link: http://subscenter.cinemast.com/he/subtitle/series/game-of-thrones/

ofir123: Fixed it!
Guessing is now only done when the search had failed.

Diaoul: Why would the search fail? Guessing logic shouldn't be necessary.

ofir123: In case the website is down, or something went wrong with the search.
I think it's a good fail-safe mechanism.

Diaoul: If the website is down, download will fail anyway. I prefer to have only one way of doing the search and if that fails that users open an issue.
Maybe it's fail-safe but it will introduce two ways of getting the URL and as a result two users with the same request could end un with different results.

Please remove this from PR and keep it somewhere so we can use it, if necessary.

ofir123: Removed!

ofir123:
Hey! I fixed all the notes, are you going to merge soon?

Diaoul:
Yep, I think it'll land in 1.1.

Issue #383, opened 2014-05-26
Add legendas.tv bigger PT-BR Subtitle Provider


felipesalomao:
Hi @Diaoul, you told that would be better develop add providers when release subliminal 0.7, and we are on 7.0.4 now :) 
Legendas.tv was totally remodeled a few months ago and does not require authentication anymore. In fact, many things changed and it's now much simpler to make the provider since the URLs are straight links and not a mess of javascript and redirections like before. Bigger problem told by @Diaoul before, was fact that would be necessary user/pass to download subtitles from legendas.tv provider, now does not need anymore.. Can download any subtitle without need user and pass.. Hope develop it..
i think that could help: https://github.com/MestreLion/legendastv

Thank you for great work !

ratoaq2:
I have started some experimental work on this. Subliminal architecture is awesome, I must say.

Legendas.tv has a really poor API/interface. The data is not exposed in a structured way and it makes a little bit harder than other providers. Other providers are much easier since they either use hashes or known ids (imdb_id or tvdb_id). Legendas.tv does not even distinguish between movies and shows and they have a lot of duplicated entries (duplicated movie entries, duplicated shows entries). 

One thing that I noticed: Subliminal scoring system does not have a field related to the "subtitle quality". Is this by design?
If I have several subtitles that matches exactly my release, then we could use some information from the provider (number of downloads, user rating, or any other field) to choose the one with best quality.

Just to give an example:
```
Subtitle 1:
Show A S01E01 720p HDTV DIMENSION.srt
Rate: 3 stars, 5 downloads
```

```
Subtitle 2:
Show A S01E01 720p HDTV DIMENSION.srt
Rate: 9 stars, 230192 downloads
```

Their score will be the same, but I would rather prefer to get Subtitle 2

Diaoul:
I will add this feature for 1.2 and update the providers accordingly.

felipesalomao:
Hi, your work to add legendas.tv provider is awesome!!! 
It is full developed (seems To be) ? 

It already was added and is available to any one ? I think not because o cant see legendastv in providers files and also actual version is 1.0.1, You said that it would be added in 1.2 version, if Yes, do You know when You will release it ?

Thanks for Great work !

Diaoul:
Fixed with 9785bc9746c24c8c28c4ed20812f103a070ca357

PR #513, opened 2015-09-26
#383: LegendasTvProvider implementation

ratoaq2:
This is an alternative implementation of LegendasTvProvider. This implementation follows subliminal guideline to have 1 LegendasTvSubtitle = 1 Subtitle release. The drawback (as discussed before) is that for each candidate a rar/zip file needs to be downloaded in order to extract the filenames from it. Cache is in place to avoid repetitive downloads. The advantage is accuracy: with the release names 100% accurate, subliminal can properly score the subtitles.

I decided to create a 2nd pull request since this approach is different from the other one. If you think this is a better approach, then you could close the other PR.


ratoaq2:
I should not change the resolution from 1080p to 720p. It's up to the scoring system to take that into consideration. I believe the resolution gives less scoring points than the release group and release souce.

.travis.yml
```diff
@@ -1,4 +1,4 @@
-sudo: false
+sudo: required
```
Diaoul: This is too bad :(

.travis.yml
```diff
@@ -33,6 +33,10 @@ cache:
 before_cache:
   - rm -f $HOME/.cache/pip/log/debug.log
 
+before_install:
+  - sudo apt-get update -qq
+  - sudo apt-get install -y unrar
+
```
Diaoul: Unrar is quite a strong dependency, how does subliminal behaves with legendastv provider if unrar is missing?

setup.py
```diff
@@ -29,7 +29,7 @@ def run_tests(self):
 # requirements
 install_requirements = ['guessit>=0.9.1', 'babelfish>=0.5.2', 'enzyme>=0.4.1', 'beautifulsoup4>=4.2.0',
                         'requests>=2.0', 'click>=4.0', 'dogpile.cache>=0.5.4', 'stevedore>=1.0.0',
-                        'chardet>=2.3.0', 'pysrt>=1.0.1', 'six>=1.9.0']
+                        'chardet>=2.3.0', 'pysrt>=1.0.1', 'six>=1.9.0', 'rarfile>=2.6']
```
Diaoul: Did you test with previous version of rarfiles? The requirements are for api-level compatibility with subliminal.

subliminal/providers/legendastv.py
```diff
@@ -0,0 +1,466 @@
...
+        self.no_downloads = no_downloads
+        self.rating = rating
+        self.featured = featured
+        self.multiple_episodes = multiple_episodes
+        self.timestamp = timestamp
+        self.guess = guess  # Do not need to guess it again if it was guessed before
+
+    @property
+    def id(self):
+        return '%s (%s)' % (self.name, self.subtitle_id)
```
Diaoul: Is the name required for unicity?

subliminal/providers/legendastv.py
```diff
@@ -0,0 +1,466 @@
...
+        self.imdb_id = imdb_id
+        self.type = type
+        self.season = season
+        self.year = year
+        self.no_downloads = no_downloads
+        self.rating = rating
+        self.featured = featured
+        self.multiple_episodes = multiple_episodes
+        self.timestamp = timestamp
+        self.guess = guess  # Do not need to guess it again if it was guessed before
```
Diaoul: I don't like the idea of carrying the guess around as it can be quite big and duplicates information (at the cost of CPU time, that's true)

subliminal/providers/legendastv.py
```diff
@@ -0,0 +1,466 @@
...
+            if not self.name_matches(expected.get('series'), actual.get('series')):
+                return False
+            if expected.get('season') != actual.get('season'):
+                return False
+            if not ignore_episode and expected.get('episodeNumber') != actual.get('episodeNumber'):
+                return False
+
+        return True
+
+    def name_matches(self, expected_name, actual_name):
```
Diaoul: I've added https://github.com/Diaoul/subliminal/blob/develop/subliminal/subtitle.py#L235, does that work for legendas?

subliminal/providers/legendastv.py
```diff
@@ -0,0 +1,466 @@
...
+        if expected.get('type') != actual.get('type'):
+            return False
+
+        if expected.get('type') == 'movie':
+            if not self.name_matches(expected.get('title'), actual.get('title')):
+                return False
+            if expected.get('year') != actual.get('year'):
+                if expected.get('year') and actual.get('year'):
+                    return False
+                if expected.get('title', '').lower() != actual.get('title', '').lower():
```
Diaoul: Not using name_matches here?

subliminal/providers/legendastv.py
```diff
@@ -0,0 +1,466 @@
...
+        pack_name_re = re.compile('^\(p\)')
+
+        # Regex to extract the subtitle_id from the 'href'. e.g.: /download/560014472eb4d/foo/bar
+        subtitle_href_re = re.compile('/download/(\w+)/.+')
+
+        subtitles = []
+        # loop over matched movies/shows
+        for title in titles:
+            # page_url: {server_url}/util/carrega_legendas_busca_filme/{title_id}/{language_code}
+            page_url = '%s/util/carrega_legendas_busca_filme/%s/%d' % (self.server_url, title.get('id'), language_code)
```
Diaoul: `title['id']` here, also `title` is a little bit confusing, I'd expect a title to be a string here. Maybe rename titles to candidates?

subliminal/providers/legendastv.py
```diff
@@ -0,0 +1,466 @@
...
+        if not expected_name or not actual_name:
+            return expected_name == actual_name
+
+        words = self.word_split_re.findall(expected_name)
+        name_regex_re = re.compile('(.*' + '\W+'.join(words) + '.*)', re.IGNORECASE)
+
+        return name_regex_re.match(actual_name)
+
+    @region.cache_on_arguments(expiration_time=SHOW_EXPIRATION_TIME)
+    def search_titles(self, params):
```
Diaoul: I'd prefer explicit args rather than implicit ones. Also I think you can decouple from the `params` logic for this function. Functions in general should be simple to understand, simple args, simple return value.

subliminal/providers/legendastv.py
```diff
@@ -0,0 +1,466 @@
...
+        Returns shows or movies information by querying `/legenda/sugestao` page.
+        Since the result is a list of suggestions (movies, tv shows, etc), additional filtering is required.
+        Type (movies or series), name, year and/or season are used to filter out bad suggestions.
+
+        :param params: dictionary containing the input parameters (title, series, season, episodeNumber, year)
+        :return: shows or movies information
+        :rtype: : ``list`` of ``dict``
+        """
+
+        keyword = params.get('title') if params.get('type') == 'movie' else params.get('series')
```
Diaoul: Maybe this logic can be put in the caller rather than here?

subliminal/providers/legendastv.py
```diff
@@ -0,0 +1,466 @@
...
+                item[field] = (lambda v: int(v) if v and v.isdigit() else None)(item.get(field))
+
+            # ignoring episode match since this first step is only about movie/season information
+            if self.matches(params, item, ignore_episode=True):
+                candidates.append(dict(item))
+
+        logger.debug('Titles found: %s', candidates)
+        return candidates
+
+    def query(self, language, video, params):
```
Diaoul: Same here, explicit args rather than params would be more readable and a cleaner API. The goal of the  `query` method is to be as close as possible to a query to the provider with minimal filtering (as this will occur in the scoring part anyway)

subliminal/providers/legendastv.py
```diff
@@ -0,0 +1,466 @@
...
+                    subtitle_id = (lambda m: m.group(1) if m else None)(subtitle_href_re.search(page_link))
+                    multiple_episodes = bool(div.find_parent('div', {'class': 'pack'}) or
+                                             pack_name_re.findall(link_tag.string))
+                    featured = bool(div.find_parent('div', {'class': 'destaque'}))
+                    no_downloads = (lambda v: int(v) if v and v.isdigit() else None)(
+                        (lambda m: m.group(1) if m else None)(rating_info_re.search(div.text)))
+                    rating = (lambda v: int(v) if v and v.isdigit() else None)(
+                        (lambda m: m.group(2) if m else None)(rating_info_re.search(div.text)))
+                    timestamp = (lambda d: datetime.strptime(d, '%d/%m/%Y - %H:%M') if d else None)(
+                        (lambda m: m.group(1) if m else None)(timestamp_info_re.search(div.text)))
```
Diaoul: Don't use lambdas in that sections for clarity please.

subliminal/providers/legendastv.py
```diff
@@ -0,0 +1,466 @@
...
+                        base_name = os.path.splitext(name)[0].replace('\\', '/')
+                        guess = guess_file_info(base_name + '.mkv', type=title.get('type'))
+                        if not self.matches(params, guess):
+                            continue
+
+                        subtitle = LegendasTvSubtitle(language, page_link, subtitle_id, name, guess=guess,
+                                                      imdb_id=title.get('imdb_id'), type=title.get('type'),
+                                                      season=title.get('season'), year=title.get('year'),
+                                                      no_downloads=no_downloads, rating=rating, featured=featured,
+                                                      multiple_episodes=multiple_episodes, timestamp=timestamp)
```
Diaoul: Also fill the content here, no need for an additional step.

subliminal/providers/legendastv.py
```diff
@@ -0,0 +1,466 @@
...
+            tmp.flush()
+
+            cf = RarFile(tmp.name) if is_rarfile(tmp.name) else (ZipFile(tmp.name) if is_zipfile(tmp.name) else None)
+
+            return function(cf, *args, **kwargs) if cf else None
+        finally:
+            tmp.close()
+
+    @region.cache_on_arguments(expiration_time=EPISODE_EXPIRATION_TIME)
+    def download_content(self, subtitle_id, timestamp):
```
Diaoul: You can remove the cache for this section, I want to keep caching for simple things, not actual subtitle contents.

Diaoul:
I added some comments.

You can move the re.compile in the module level if you use them in different places, otherwise just use them without compiling. See http://stackoverflow.com/questions/452104/is-it-worth-using-pythons-re-compile
Given that for the most part you use them only once, I'd go for the latter.

Diaoul:
Any updates on this?

ratoaq2:
I'm a bit busy these days, but soon I'll get back to this and implement the
suggestions and discuss the open issues.

felipesalomao:
Thank You so much @ratoaq2 ! Your work is awesome, i have been waiting years to see this become true.. And i think 95% of Job is done.. Just few changes to it work for everyone.. You as brazillian know how important and better are subtitles from legendas.tv.. Only they have 100% realible subtitles for all big TV shows and movies..

felipesalomao:
Hi, i am testing legendas.tv provider developed by @ratoaq2, it aways find subtitle.. But i am Habibs problem, dont know if is legendas.tv provider bug or sickrage subtitle module bug.. When i enable only legendas.tv provider, it not download automatically, only clicking CC on TV show page..
Then i opened log, found problem:
2015-11-26 00:06:45 FINDSUBTITLES :: Not enough services selected. At least 1 service is required to search subtitles in the background

Then i was to subtitle settings checo and legendas.tv was unchecked.. But i enabled it and saved settings.. Tried yesterday and today.. Tried restart sickrage too and happened again.. It dont keep legebdas.tv subtitle provider enabled.. After some time it disable for no reason.. Do You know how fix it ?

fernandog:
@ratoaq2 how do you handle this?

Only one file for all releases?
legendas. tv /download/564f756c88901/Major_Crimes/Major_Crimes_S04E13_HDTVx264_LOL_MSD_RMTeam_FUM_AFG_DIMENSION

fernandog:
@ratoaq2 I pass min_score = 90 and 100 but ltv/subliminal is still downloading subs for:
MF.S07E07.Phil's.S*,.S*.H*.1080p.WEB-DL.DD5.1.H.264-Oosh

but the only releases availables are:
M.F.S07E07.HDTV.x264-FLEET
M.F.S07E07.INTERNAL.HDTV.x264-BATV
M.F.S07E07.720p.HDTV.x264-FLEET
M.F.S07E07.INTERNAL.720p.HDTV.x264-BATV
M.F.S07E07.XviD-AFG
M.F.S07E07.INTERNAL.XviD-AFG
M.F.S07E07.480p.x264-mSD

@Diaoul score = 100 is exact filename match?

Diaoul:
Nope, score is based on various matches, 100 is all matches at the same time without equivalences. Read [this](http://subliminal.readthedocs.org/en/latest/api/score.html) and [that](http://subliminal.readthedocs.org/en/latest/user/how_it_works.html#scoring).

fernandog:
@Diaoul didn't get that equations, sorry.

a srt 720p-DIMMENSION  works for 1080p-DIMMENSION, also for 420.
what matters IMO is the same release group.
so what's the minimal score for: series name + episode + season + relese group?

Diaoul:
I'm open to suggestions on how to improve the scoring. For that I need a large dataset to build unittests and proper equations.
Scores are explicited [here](https://github.com/Diaoul/subliminal/blob/master/subliminal/video.py#L141-L143) max is hash + hearing_impaired, scale it down to 100.
Ping @pannal

fernandog:
@Diaoul 
so min_core = 132 to match: series name + season + episode  + resolution + video_codec ?
to match release group just add 4 , so it will be 136 ?

why not set in the code min_core default value to 132 ? default is 0 so it will always get bad subs if the sub available are not those we want.

subliminal/providers/legendastv.py
fernandog: I can respond this one: yes.

eache download id(link) has one rar file and it contains multiple subs:

See this log:
https://gist.github.com/fernandog/7103aeea8c272395939f

If you don't add the name to it, all IDs would be 5660810a16e42, so that's why he added the name

subliminal/providers/legendastv.py
fernandog: so it's here that we need to remove country name i.e (US) or (UK), etc ?

subliminal/providers/legendastv.py
Diaoul: OK, in that case use `'%s-%s' % (self.subtitle_id, self.name)` instead please.

subliminal/providers/legendastv.py
fernandog: each subtitle result/upload has an unique ID. that ID its also the name of the rar file
560014472eb4d.rar - and that's why he added `return '%s (%s)' % (self.name, self.subtitle_id)`

So I agree he could use candidates

subliminal/providers/legendastv.py
fernandog: If I understood correct cache is a good thing for LTV
ltv first upload the most common subtitle releases then, they update the .rar file adding new subs for new releases. They don't create new subs.
So cache is a good thing here to avoid keep re-downloading the same .rar and nothing has changed.

fernandog:
@Diaoul I tried to reply some of your comments, don't know if they are helpful

felipesalomao:
@ratoaq2 
Hi, i was talking with @fernandog 
On this script, it fail to download some subtitles because some legendas.tv legendary put name of file in this format: Z.Nation.S02E15.720P.HDTV.x264-KILLERS[EETV].srt. 
Problem that algorithm don't understand that KILLERS[EETV] is release KILLERS, then it don't have enough score to download it.. On script, could ignore if have something inside "[]" characters.. Understood what I mean?! Then would aways download all subs every time.. I am downloading all my TV shows using this script on Sickrage for more than one month (download about 10 episodes per week) and unique time that it wasn't downloaded was because of this problem.. I aways look on log..

See pictures above.. If need anything, just let me know! Thanks for your great work!!!
Merry Christmas and Happy new year ;)
image
image

fernandog:
@Diaoul @ratoaq2 we did those changes to legendastv.py

https://github.com/SickRage/SickRage/commit/7c616f0cfeed08fe3b1abc54206456fe0723fb8b

https://github.com/SickRage/SickRage/commit/1b1a10c7a772016969fbb3a3532ee9086ffbde4f

https://github.com/SickRage/SickRage/commit/aef7a81a7c7941901afdf8dc5a7a2cd084c62f77

subliminal/providers/legendastv.py
Zopieux: subliminal stores its cache into `~/.config`, which is a bad idea because it should be used for **permanent** user **configuration** files. Ever-growing, **safely deletable** cache directories should be stored in `~/.cache` instead.

Diaoul: Please report any issue on the bug tracker not here. Subliminal is a lib, it doesn't store anything in `.config`. Its CLI does if that's what you're using.

Zopieux: It's not a subliminal issue. I just wanted to highlight this fact to @fernandog so he does not use `.config` to store cached files.

Diaoul: There is nothing to highlight to @ratoaq2, the fact that it uses .config to store cache file is configured by the CLI and has nothing to do with @ratoaq2's PR.

.travis.yml
ratoaq2: I agree. As you may know unrar has license restrictions. I'm not aware if there's another option to avoid the unrar utility/library dependency.

https://rarfile.readthedocs.org/en/latest/faq.html

.travis.yml
ratoaq2: If the unrar is missing the provider will still work for any file that's not a rar file. Most of the subtitles in legendastv are rar files but there's still some zip files there. After downloading the file, the provider will check if it's a rar file (this will pass even if unrar utility is not present) and if it's a rar file, it will try to extract it. If there's no unrar utility installed, then you'll get the following exception:

`RarCannotExec: Unrar not installed? (rarfile.UNRAR_TOOL='unrar')`

setup.py
ratoaq2: I'll check what's the required minimal version and use it. I know that the version 2.7 or 2.8 (I need to check) allows me to extract the file without using a temporary file, but I would rather stick with lower version dependency to increase compatibility.

subliminal/providers/legendastv.py
ratoaq2: I'll remove this field and it will be computed again

subliminal/providers/legendastv.py
ratoaq2: I believe that you're refering to sanitized_string_equal method.

Legendastv database is not consistent and there are cases where you search for a movie/series and the result is something completely different (e.g.: Searching for "The Big Bang Theory" will return, among the correct results, a completely wrong one: "Nanny McPhee Returns")
The name_matches will just filter out bad results, the result name should contain all words in the order of appearance (e.g.: given the search is "The Big Bang Theory" and the result list is [The Big Bang Theory S01, The Big Bang Theory S02, Nanny Mcphee Returns], the filtered list will be [The Big Bang Theory S01, The Big Bang Theory S02].

 sanitized_string_equal will help me here:
`if expected.get('title', '').lower() != actual.get('title', '').lower():`

subliminal/providers/legendastv.py
ratoaq2: For certain movies, there's no year information in legendastv database, but there's an exact name match, therefore this result entry shouldn't be filtered out. name_matches does a different job, but now I can make use of `sanitized_string_equal`.

ratoaq2: @fernandog 
All these title manipulations shouldn't be done in subliminal, they should come from guessit.

subliminal/providers/legendastv.py
ratoaq2: Agreed. I'll rename it to `candidates` and `search_candidates`

subliminal/providers/legendastv.py
ratoaq2: Extracting all can lead to memory issues since we're downloading all candidates (15-20 files) and each file can have subtitles for several releases or several episodes. Just a few KB can grow to tens of MB easily. Maybe it's better to keep the binary data and the extraction will only happen if that subtitle was the chosen one.

subliminal/providers/legendastv.py
ratoaq2: STATUS: [Changed] to `'%s-%s' % (self.subtitle_id, self.name.lower())`

subliminal/providers/legendastv.py
ratoaq2: STATUS: [Changed] Using explicit arguments instead of params.

subliminal/providers/legendastv.py
ratoaq2: STATUS: [Changed]: Fixed

subliminal/providers/legendastv.py
ratoaq2: STATUS: [Changed]: Lambdas removed

subliminal/providers/legendastv.py
ratoaq2: STATUS: [Changed]: No cache is used anymore

ratoaq2:
Finally I had time to come back to this.
Please find my comments on each issue. In the meanwhile I'll make sure the tests are green in all different environments.

fernandog:
@ratoaq2 please before update this PR check our changes in SickRage. I tagged you there.

ratoaq2:
In my fork, the build got green, but here it got stuck.

.travis.yml
Diaoul: Should we disable the provider if unrar isn't installed? Or does it give good results even without unrar?

.travis.yml
Diaoul: Is the cache still supported with sudo required? https://github.com/Diaoul/subliminal/pull/513/files#diff-354f30a63fb0907d4ad57269548329e3L29

subliminal/providers/legendastv.py
```diff
@@ -0,0 +1,486 @@
...
+        return '%s-%s' % (self.subtitle_id, self.name.lower())
+
+    def get_matches(self, video, hearing_impaired=False):
+        matches = super(LegendasTvSubtitle, self).get_matches(video, hearing_impaired=hearing_impaired)
+
+        # The best available information about a subtitle is its name. Using guessit to parse it.
+        guess = guessit(self.name, {'type': self.type})
+        # Cleaning release_group: foo[bar] -> foo
+        if guess.get('release_group'):
+            guess['release_group'] = re.sub('\[\w+\]', '', guess['release_group'])
```
Diaoul: Should that be extended to other providers? Or is it only legendastv related?

ratoaq2: Yes, I think so. I was expecting this issue to be handled by guessit, but I
got a reply from them saying that this part will not change and we need to
do post-processing:
https://github.com/guessit-io/guessit/issues/196#issuecomment-177627695

fernandog: Do you think is worth to compare relesegroup with/without the brackets and get the highest score when matching both?

.travis.yml
ratoaq2: I strongly believe that we should disable the provider, but I'll need to check it later.
With the current implementation, as soon as you get a rar file an exception will be raised and no subtitle will be downloaded. If we decide to not disable the provider then I'll need to change the logic a little bit in order to proceed with the other possible candidates.

fernandog: at SickRage we have unrar2 in our /lib folder. Would it work?

subliminal/providers/legendastv.py
Diaoul: I meant can you get rid of `name_matches` and use `sanitized_string_equal` instead? For your examples you can just `s/[SE]\d{2}//g` and send the result to `sanitized_string_equal`. I'd prefer have only one function that does the string match, if need be it can be improved to fit your use case.
I may need to add parameterized unittests to `sanitized_string_equal`

subliminal/providers/legendastv.py
```diff
@@ -0,0 +1,486 @@
...
+                                                      multiple_episodes=multiple_episodes, timestamp=timestamp)
+
+                        logger.debug('Found subtitle %s', subtitle)
+                        subtitles.append(subtitle)
+
+                next_page_link = soup.find('a', attrs={'class': 'load_more'}, text='carregar mais')
+                page_url = self.server_url + next_page_link['href'] if next_page_link else None
+
+        # High quality subtitles should have higher precedence when their scores are equal.
+        subtitles.sort(key=lambda s: (s.featured, s.no_downloads, s.rating, s.multiple_episodes), reverse=True)
```
Diaoul: This will be properly implemented in the future, you can leave as is for now.

subliminal/providers/legendastv.py
ratoaq2: It was just an example. Another example would be:
Searching for "Superman" and the results could be ["Superman", "Superman: O Filme", ...] (and in this case both should be considered as candidates)

I could try to use `sanitized_string` on both strings and then test if one startswith the other one. It would be something like `sanitized_string_compare`or `sanitized_string_startswith`

But first I'll run some tests and check the results.

subliminal/providers/legendastv.py
```diff
@@ -0,0 +1,486 @@
...
+        :param content: the downloaded binary content (rar/zip)
+        :param subtitle_name: the filename to be extracted
+        :return: the subtitle content
+        :rtype : ``string``
+        """
+        return self._uncompress(content, lambda cf, name: fix_line_ending(cf.read(name)), subtitle_name)
+
+    def _uncompress(self, content, function, *args, **kwargs):
+        # Download content might be a rar file (most common) or a zip.
+        # Unfortunately, rarfile module only works with files (no in-memory streams)
```
Diaoul: In the docs indeed there's no mention of in-memory streams however any file-like object should work. At least it's what I understand from the [source code](https://github.com/markokr/rarfile/blob/master/rarfile.py#L1747-L1771)

ratoaq2: You can check the history:
https://github.com/markokr/rarfile/commits/master/rarfile.py

This was introduced on 20.11.2013 for version 2.7 by this commit: https://github.com/markokr/rarfile/commit/3ed04e372f472464634625d234efb8e56677d5ad#diff-5689b095d6dc734b474f99921552f98c

I could get rid of the temporary files, but the rarfile dependency should be changed from rarfile>=2.4 to rarfile>=2.7

.travis.yml
Diaoul: OK, maybe we can disable it the check method?

subliminal/providers/legendastv.py
Diaoul: Maybe add a `allow_partial` flag to `sanitized_string_equal` or something so it fits your usecase, I really want to centralize the way titles are compared.

subliminal/providers/legendastv.py
ratoaq2: @Diaoul I haven't changed this. Should we wrap all calls to guessit(string) and put this logic there?
@fernandog I don't see any benefit. ABC[xyz] and ABC represents the very same release.

fernandog: yes but some uploaders add the brackers (wrong). just a fail safe

ratoaq2:
I believe I have addressed all the issues, but I haven't moved the `release_group` post-processing to a common place. There's no sudo required in travis config anymore (fortunately unrar is in its [whitelist](https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-precise). Moved to rarfile 2.7 in order to use in-memory streams (it also helps that in windows there's no need to have unrar.exe on the system path)

I could fix the docstrings to have the correct data type to follow the subliminal standards and I could squash the commits to a single commit as well. Just waiting to know whether other changes will be required before doing that.

fernandog:
@ratoaq2
Arent you going to handle connection errors when 
ltv is down or 503 or undermaintance?

Enabling subliminal logs It shows a traceback:

A2015-12-19 10:34:35 ERROR    Unexpected error in provider 'legendastv', discarding it

fernandog:
https://github.com/SickRage/sickrage-issues/issues/267

Diaoul:
@fernandog: it's the expected behavior. Providers can raise anything related to network errors, the high level API takes care of this by discarding the provider. If a network failure occurred it's likely that it will happen again for the next query so discarding it is the proper response.

fernandog:
@Diaoul  isn't better to show some other message other then "no results found" when network errors ? something like "provider not available at the moment"

subliminal/providers/legendastv.py
Diaoul: This is not an issue related to legendastv, please open a separate issue so we can discuss it here. Also if you have examples in various providers that will prove very useful.

subliminal/subtitle.py
```diff
@@ -191,16 +191,26 @@ def sanitize_string(string, replacement=''):
     return re.sub('[^ a-zA-Z0-9]', replacement, string)
 
 
-def sanitized_string_equal(string1, string2):
+def sanitized_string_equal(string1, string2, allow_partial=False):
     """Test two strings for equality case insensitively and ignoring special characters.
 
     :param str string1: the first string to compare.
     :param str string2: the second string to compare.
+    :param bool allow_partial: the second string to compare.
```
Diaoul: Hasty copy paste here :wink:

tests/test_legendastv.py
```diff
@@ -0,0 +1,321 @@
...
+        'Man of Steel (BDRip.x264-Larceny Subrip).srt (527d55c0c02bc)',
+        'Man of Steel (2013) DVDRip XviD-MAXSPEED.srt (54edd551e869a)'
+    }
+
+    with LegendasTvProvider(USERNAME, PASSWORD) as provider:
+        provider.initialize()
+        subtitles = provider.query(language, title=video.title, year=video.year)
+        provider.terminate()
+
+    assert {'%s (%s)' % (subtitle.name, subtitle.subtitle_id) for subtitle in subtitles} == expected_subtitles
```
Diaoul: You can use tuples here, it's hashable:

    ('Superman Man Of Steel 2013 ANOTHER NEW SOURCE TS XViD UNiQUE.srt', 'a50ef54a73c490f3d7f63c333f5d3e07')

Diaoul:
@fernandog I'm not sure I understand here. You want me to add a `print` somewhere? There's already a log message and the information is available in [discarded_providers](https://github.com/Diaoul/subliminal/blob/master/subliminal/api.py#L46).
Please do not use this PR for unrelated questions.

subliminal/providers/legendastv.py
ratoaq2: An issue #574 was opened for it  Should I leave this piece of code here? Or remove it?

Diaoul: Thanks, keep it for now. I'll fix this for every provider after merge.

ratoaq2:
Done. I have fixed the remaining 2 issues.

On a small note, during some manual tests I spotted a small issue when using sanitized_string_equal:
The X-Files Season 10 (as in legendas.tv website) trying to match The X Files S10E03:

```sanitize_string1 = 'the x files'```
```sanitize_string2 = 'the xfiles'```

**Comment by Diaoul** [comment]
What about replacing dashes with spaces? It makes sense.

**Comment by ratoaq2** [comment]
This is the way that I have fixed it in my local env, I changed sanitize_string method:

```
def sanitize_string(string, replacement=''):
    """Replace any special characters from a string.

    :param str string: the string to sanitize.
    :param str replacement: the replacement for special characters.
    :return: the sanitized string.
    :rtype: str

    """
    return re.sub('[^ a-zA-Z0-9]', replacement, string.replace('-', ' '))
```

Or should I fixed it the provider itself?

Diaoul:
I have a fix pending with a less brutal approach. The current `sanitize_string` can rip off more than just special characters, it'll also remove accentuated characters, this is bad.
Leave it that way for now, I'll make the fix in develop.

felipesalomao:
I tired download this: 
dc's legends of tomorrow 
Problem that it not found subtitle because name is little different:
DCs.Legends.of.Tomorrow.S01E02.1080p.HDTV.X264-DIMENSION
Problem was on '

Could ltv module handle this?

fernandog:
There is ' and `

Diaoul:
I think some special characters should be replaced with a space (like `-` or `:`).
Some other should be removed (like `'`).
And of course replace multiple spaces with one.

Thoughts?

Diaoul:
@felipesalomao: there is no way for subliminal to know that DCs is actually DC's so any search will be made with DCs which may not return the correct result.
I think SickRage (or whatever you use) should input the correct series name to subliminal before issuing a search.

ratoaq2:
@Diaoul I agree 100%: `-` and `:` to space; `'` to be removed; multiple spaces to 1

About @felipesalomao issue: I'm still not 100% sure if something needs to be done in the provider. I know that [search_candidates](https://github.com/Diaoul/subliminal/pull/513/files#diff-e33b1676e0aad8bc5e406d5983fe5bc7R173) will not find anything if you search [`DC's`](http://legendas.tv/legenda/sugestao/DC's) but it will find correct results if we use [`DCs`](http://legendas.tv/legenda/sugestao/DCs). On the other hand, if you search [`Marvels Jessica Jones`](http://legendas.tv/legenda/sugestao/Marvels%20Jessica%20Jones) it will give you 0 results, but if we use [`Marvel's Jessica Jones`](http://legendas.tv/legenda/sugestao/Marvel's%20Jessica%20Jones) it will return the correct ones. This issue doesn't happen with `-`. I'll leave that for now.

ratoaq2:
Trying to keep the PR updated.
- I have rebased and squash'ed the commits (develop is broken, there's a quick fix until it get green again)
- I reviewed the docstrings (only small changes/fixes).
- Added cache for the subtitle_names in order to avoid downloading all candidates every time you execute subliminal for a specific file. The cache key is `subtitle_id` and `timestamp`; and the cached value is an array of strings (no binary content is cached)

No other change.

ratoaq2:
- Rebased and moved the release_group strip logic to a common place.

fernandog:
@ratoaq2 seems ltv changed something:

HTTPError: 403 Client Error: Forbidden for url: http://f.legendas.tv//

full logs:
https://gist.github.com/fernandog/da81f24f234bf577442f

ratoaq2:
@fernandog I tested several times this morning and during the afternoon. I couldn't reproduce this issue and it's really strange that the url is different from the one that's in the code. Probably there was a site maintenance and that was a redirect url during that time, but hard to guess.

@Diaoul Let me know if there's anything I should change/refactor/remove. I'm not working on it anymore.

fernandog:
@ratoaq2 
nah. still don't working here: 
it works for one search, then fail to others.
can you join irc?

ratoaq2:
- Rebased and squash'ed commits.
- Added User-Agent

felipesalomao:
Hi, have some estimated time for provider full implementation? Thank you guys

Diaoul:
I've finished up with other 2.0 changes so I just need to merge this as is and eventually I will come up with some changes to it.
2 others providers PRs are also in this case.

ratoaq2:
- Merged and rebased. No changes.

Diaoul:
Based on your work and to have a clean history I added legendas.tv provider with 9785bc9746c24c8c28c4ed20812f103a070ca357. I initially merged your branch and reworked the caching part for archives to avoid downloading twice the same archive but ended up changing a lot of things around.
Please let me know if it works properly for you.

Thanks for the PR :+1:

ratoaq2:
Great.
I'll be testing it during the following couple of days and I'll report back to you.

PR #520, opened 2015-10-03
Get enconding from OpenSubtitles provider

mouchar:
OpenSubtitles returns 'SubEncoding' that we can use to avoid guess_encoding()

### DIFF
--- a/subliminal/providers/opensubtitles.py
+++ b/subliminal/providers/opensubtitles.py
@@ -23,8 +23,8 @@ class OpenSubtitlesSubtitle(Subtitle):
     series_re = re.compile('^"(?P<series_name>.*)" (?P<series_title>.*)$')
 
     def __init__(self, language, hearing_impaired, page_link, subtitle_id, matched_by, movie_kind, hash, movie_name,
-                 movie_release_name, movie_year, movie_imdb_id, series_season, series_episode):
-        super(OpenSubtitlesSubtitle, self).__init__(language, hearing_impaired, page_link)
+                 movie_release_name, movie_year, movie_imdb_id, series_season, series_episode, encoding):
+        super(OpenSubtitlesSubtitle, self).__init__(language, hearing_impaired, page_link, encoding)
         self.subtitle_id = subtitle_id
         self.matched_by = matched_by
         self.movie_kind = movie_kind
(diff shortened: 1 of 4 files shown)


subliminal/providers/opensubtitles.py
```diff
@@ -23,7 +23,7 @@ class OpenSubtitlesSubtitle(Subtitle):
     series_re = re.compile('^"(?P<series_name>.*)" (?P<series_title>.*)$')
 
     def __init__(self, language, hearing_impaired, page_link, subtitle_id, matched_by, movie_kind, hash, movie_name,
-                 movie_release_name, movie_year, movie_imdb_id, series_season, series_episode):
+                 movie_release_name, movie_year, movie_imdb_id, series_season, series_episode, encoding=None):
```
Diaoul: No need to make it optional

subliminal/providers/opensubtitles.py
```diff
@@ -35,6 +35,7 @@ def __init__(self, language, hearing_impaired, page_link, subtitle_id, matched_b
         self.movie_imdb_id = movie_imdb_id
         self.series_season = series_season
         self.series_episode = series_episode
+        self.encoding = encoding
```
Diaoul: `encoding` is a property of the parent `Subtitle` class, put it in the super call and edit the parent class so the constructor accepts the encoding argument.

Diaoul:
Thanks for the PR, I added a few comments.

mouchar:
Antoine,
I changed my patch according to your proposals.
I had to put try-except block around ```self.content.decode``` because OpenSubtitles API returns garbage in SubEncoding for specific subtitles (eg. IDSubtitleFile=1954443141 has ```SubEncoding='CP1252|ISO-8859-'```)
Added simple condition on encoding initialization to one test.

skorokithakis:
Can this get merged? I'm getting some errors where subliminal thinks the encoding is Latin (even though `charade` says 0.99 Greek (it is Greek) and I'm getting corrupted UTF8 output. This would fix that.

Diaoul:
Sorry I thought I commented that PR already.

Please filter out invalid encodings when setting the property rather than the try/catch block on the text property. The encoding property being part of the public API I don't want garbage there.

mouchar:
Hi, ```encoding``` attribute contains the original value as it came from from the OS provider. 
I can't sanitize its value within the constructor, as I need to have ```content``` available (to ensure that content is decodable).
The most I could to is to update ```self.encoding``` value according to output of guess_encoding(), in case that self.content.decode(self.encoding) fails. It would fix the value of encoding attribute when accessing ```text``` property. But I'm not much happy with this either as it would bring strange side-effect to Subtitle.text property.
If you're not satisfied with proposed solution, feel free to close this PR as I'm out of other ideas, sorry :-)

Diaoul:
It's no big deal if the encoding is the wrong one, that's still better than nothing. What I mean is check at least it is a valid codec with something like https://docs.python.org/2/library/codecs.html#codecs.getdecoder or similar when setting the property. Even better, try to split using `|` and get the first one if this is common in OpenSubtitles

mouchar:
@Diaoul I've just updated my PR. As you can see, I used ```codecs.CodecInfo``` object to:
  1. ensure the encoding provided is supported by Python
  2. get proper codec name, effectively translating all aliases and making case conversions (e.g. 'WinDows-1250' == 'windows_1250' == 'CP1250' == 'cp1250')

So Subtitle.encoding now gets either canonical name of supported codec, or None if input value is empty, unsupported or garbled beyond repair.
I intentionally don't attempt to parse ```SubEncoding``` because its value randomly depends on how much broken is OpenSubtitles database ;) Method ```guess_encoding()``` is more reliable in this case.

Diaoul:
LGTM, thanks!

PR #529, opened 2015-10-20
Use regex to find external subtitles

getzze:
In the search_external_subtitles function, only subtitles of the form <video_name>.eng.srt. With also some errors, like  <video_name>[eng].srt is matched to `eng]`.
Using a regex allow for a better match, of the form en, eng, en_GB or eng_GB.

### DIFF
--- a/subliminal/video.py
+++ b/subliminal/video.py
@@ -255,13 +255,12 @@ def search_external_subtitles(path):
 
         # extract the potential language code
         language_lookup = os.path.splitext(p)[0][len(fileroot):].replace(fileext, '')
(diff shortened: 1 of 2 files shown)


Diaoul:
Please add unitttests

getzze:
Ok, I changed the regex to finally pass all the tests and I added two new tests. Sorry I am not used to testing, although it is a very good practice!

Diaoul:
Are media players able to read subtitles like that? What's the use case?

Diaoul:
Say you have a movie in french `Something.2015.[FR].mkv` with `Something.2015.[FR].srt`, how does that behave?

PR #532, opened 2015-10-30
Fix napiprojekt tests

bogdal:
This PR fixes tests. Now we use real ones hashes.

Discussion: https://github.com/Diaoul/subliminal/pull/489#issuecomment-152348534


Diaoul:
Thanks for the quick fix

PR #562, opened 2016-01-25
Changed all log messages to proper formatting.

ofir123:
Log messages would cause errors in Python 3.5 because of bad formatting, so I've fixed all the log message lines.
Error screenshot:
image

### DIFF
--- a/subliminal/api.py
+++ b/subliminal/api.py
@@ -139,7 +139,7 @@ class ProviderPool(object):
         if name not in self.providers:
             raise KeyError
         if name not in self.initialized_providers:
-            logger.info('Initializing provider %s', name)
+            logger.info('Initializing provider %s' % name)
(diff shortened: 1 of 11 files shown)


Diaoul:
The fix is wrong, the formatting should be done by the logger functions for performance optimization. I don't now why you have this error, maybe changing %r to %s will solve the issue? Can you try that?

ofir123:
Still an issue.. even after changing to %s:
image

ofir123:
Since most of the time is spent on HTTP access to the subtitle sites, is it really necessary to optimize log messages?

Diaoul:
It's the way it is supposed to work see the [official doccumentation](https://docs.python.org/3/library/logging.html#logging.Logger.debug). I'll try to reproduce your issue but I think there is something wrong with your setup.

Diaoul:
I'm closing since this is a local issue.

Issue #563, opened 2016-01-26
Adic7ed provider failed with x-files series


mac1202:
Subliminal failed to find subtitle for the new x-files episodes. Here the error i got

    INFO:subliminal.api:Listing subtitles with provider 'addic7ed' and languages {<Language [fr]>}
    INFO:subliminal.api:Initializing provider addic7ed
    INFO:subliminal.providers.addic7ed:Logging in
    DEBUG:subliminal.providers.addic7ed:Logged in
    ERROR:subliminal.api:Unexpected error in provider 'addic7ed', discarding it
    Traceback (most recent call last):
      File "/usr/lib/python3.5/site-packages/subliminal/api.py", line 116, in list_subtitles
        provider_subtitles = self[name].list_subtitles(video, provider_languages)
      File "/usr/lib/python3.5/site-packages/subliminal/providers/addic7ed.py", line 259, in list_subtitles
        return [s for s in self.query(video.series, video.season, video.year)
      File "/usr/lib/python3.5/site-packages/subliminal/providers/addic7ed.py", line 214, in query
        show_id = self.get_show_id(series, year, country)
      File "/usr/lib/python3.5/site-packages/subliminal/providers/addic7ed.py", line 187, in get_show_id
        show_ids = self._get_show_ids()
      File "/usr/lib/python3.5/site-packages/dogpile/cache/region.py", line 1053, in decorate
        should_cache_fn)
      File "/usr/lib/python3.5/site-packages/dogpile/cache/region.py", line 657, in get_or_create
        async_creator) as value:
      File "/usr/lib/python3.5/site-packages/dogpile/core/dogpile.py", line 158, in __enter__
        return self._enter()
      File "/usr/lib/python3.5/site-packages/dogpile/core/dogpile.py", line 91, in _enter
        value = value_fn()
      File "/usr/lib/python3.5/site-packages/dogpile/cache/region.py", line 610, in get_value
        value = self.backend.get(key)
      File "/usr/lib/python3.5/site-packages/dogpile/cache/backends/file.py", line 220, in get
        with self._dbm_file(False) as dbm:
      File "/usr/lib/python3.5/contextlib.py", line 59, in __enter__
        return next(self.gen)
      File "/usr/lib/python3.5/site-packages/dogpile/cache/backends/file.py", line 215, in _dbm_file
        "w" if write else "r")
      File "/usr/lib/python3.5/dbm/__init__.py", line 88, in open
        raise error[0]("db type could not be determined")
[...]

PR #566, opened 2016-01-27
Added dash character to series name regex in provider addic7ed

kakai248:
Fixes this bug: https://github.com/Diaoul/subliminal/issues/563#issuecomment-174997560


Diaoul:
Done with unittests in d125ffa

Issue #416, opened 2014-10-26
add support for rar archives


djbijo:
is it possible to scan for subtitles for videos that are inside archives?
as xbmc supports rar there is no need to extract archives so it seems logical that subliminal will work with archives.
i'm not a python coder but i ran into rarfile https://rarfile.readthedocs.org/en/latest/api.html , maybe this is the answer? or even search by file/dirname and skip the embedded sub check at this case.

Diaoul:
So about this, why not extract the RAR? Compression ratio is bad anyway on videos isn't it?
This seems like a really custom need.

renini:
Its only rarred because it is split in mulitple files for downloading reasons, no compression is being used.

I'd like to keep them as i got them.

brp-david:
I'd like to chime in on this one too. I play my content via a media server that of course handles playing videos from packed files and I need to keep the rars around for torrent seeding either way so it's very redundant to unpack the files. Today I trick subliminal to download with this oneliner: 
`find <tvshows-path> -iname "*.rar" -type f -mmin -720 -not -iname "*sample*" -print0 | sed -e 's/rar/mkv/g' | xargs -0 subliminal download -v -l en`

This however has the issue of subliminal not being able to detect that the subtitle already exists and it gets downloaded multiple times. Subliminal being able to detect videos in rars would make my life complete!

Diaoul:
We can close this now

PR #576, opened 2016-02-06
Support for scanning archives for video files

brp-david:
#416 Added the possibility to include archives (rar) in scan. Files in the archive are listed via the rarfile-module and then scanned like they would be normal files. Hashes and subtitle/audio-track information is of course skipped so it has it's limitations.

First abstracted the archive handling more but since I don't think there will be any requests to support anything but rar-files (since they're the predominant scene-release format) it didn't really feel like it was worth the effort. 

3 test cases are failing for me both before and after my changes so I'm guessing it's platform issues: 
- doctest (docs/user/usage.rst:120) - Seems to be encoding issue. 
- test_video_exists_age (test_video.py:34) - Timing issue? There's only a 1 second window to get the time "right". Fails sometimes, sometimes not, but maybe not a problem where the tests run.
- test_scan_videos (test_video.py:341) - 'LocalPath' object has no attribute 'mksymlinkto', can't get this one to go through on any of the py versions I've tried on windows or freebsd.


Diaoul:
Thanks! That looks like a really nice PR.
I'm not a big fan of the `_scan_videos_in_path` part, I do get the need for it but maybe I'll think of something more "API-friendly".

Diaoul:
Why the recursion? You can just call `scan_video` for each filename in `rar_file.namelist()`?
Sorry if I'm missing something obvious but nested rarfiles isn't really a thing, is it?

brp-david:
For me it just felt neater that way since you're doing the same things, you'd have to ceck if it's a video, check if it's a hidden file. Not much but still duplicate checks. I'd be perfectly fine with changing it if you prefer though!

brp-david:
Just saw that `scan_video` checks that it is a video file anyway, so the only check you might want to duplicate is the "hidden-file"-one but could just skip that one for archive files.

brp-david:
Simplified the scanning, removed the recursive call and just scanned each file. Handling the ValueError which will be raised if it's not a video-file and just continuing if caught.

subliminal/cli.py
```diff
@@ -271,11 +271,12 @@ def cache(ctx, clear_subliminal):
 @click.option('-m', '--min-score', type=click.IntRange(0, 100), default=0, help='Minimum score for a subtitle '
               'to be downloaded (0 to 100).')
 @click.option('-w', '--max-workers', type=click.IntRange(1, 50), default=None, help='Maximum number of threads to use.')
+@click.option('--scan-archives', is_flag=True, default=False, help='Scan archives for videos (Supported formats: Rar)')
```
Diaoul: Can you think of a short version of this options? Nothing comes to mind for me.

brp-david: I got stumped myself, that's why I left it like that :p Maybe capital A or r/R even though that's usually recursive

ratoaq2: `-sa` :+1:

brp-david: I don't really like the double letter flags (but that's just me). But to me -sa could also mean -s + -a, I tend to write the flags together when I can.

Diaoul: Letter `z` is usually associated with compression even if that's for `zip`, I think it might be appropriate as the flag is for archives in general (we could support other formats in the future). Otherwise the uppercase `A` is OK too.

brp-david: I like z! Even though it feels bad to break the pattern of short version resembling the long version...

brp-david:
Added the flag. I also vow to not commit stuff (and especially push) when I'm not using my proper development environment. "It's just a small change! What could possibly go wrong!". Famous last words.

brp-david:
@Diaoul, anything else you'd like me to adjust here? Merged develop and there's some failing tests but that seems to be from outside my commits scope.

Diaoul:
Actually I was thinking that `scan_video` is made for existing videos and since nothing from there applies to an archive except the `Video.fromguess` I wanted to handle that directly in the `scan_videos` function. It feels wrong to have `scan_video` skip all the stuff based on a `is_archive` flag, kinda changes the purpose of the function.

Also I'm thinking moving the metadata refining stuff (with enzyme) in its own refiner. I don't know yet how to compose with refiners (should I split refiners for movies and refiners for episodes? Should I create an omdb_episode refiner and keep a single refiner list such as ['metadata', 'omdb', 'tvdb', 'omdb_series'])
I don't want refiners to be too complex so I don't want configurable refiners.

I'll merge when I figure out answers to those questions, ideas are welcome :)

brp-david:
I think the scan_video/scan_videos-question is very valid, and since it's basically just 2 rows that needs to be moved out I'd say that's a better solution. (the fromguess and external subtitle-check). I didn't realize that it was that little left. 

As for the refiners I don't have enough understanding to give any opinions.

Diaoul:
I'm merging this with latest develop. Just a quick question, are media players able to play the video with subtitles when:

1. subtitles exist next to the archive?
2. subtitles exist in the archive?

Thanks

Diaoul:
Is it common to have multiple video files in an archive?

brp-david:
Not at all, that could definitely be simplified aswell. It should always be only one video and basically nothing else. Everything else in a release should be outside.

brp-david:
As to the subitles next to the archive/in the archive I've been very unsure of this myself. I could test this a bit. Also unsure if it's the name of the archive or the file in the archive that is used by mediaplayers.

brp-david:
A quick test shows that subtitles in the archive isn't really a thing. Neither VLC nor Kodi handles this. 
As for subtitles beside the archive Kodi handles this if it has the same name as the video filename in the archive.

Diaoul:
OK thanks for the tests. I've merged this in my local branch and will add a unit test for the external subtitles.
I've also done a little refactoring for the sake of separation of concerns. Tests are green so I should be able to push soon.

Diaoul:
Can you test the develop branch?

brp-david:
Excellent! I tested on my box and works nicely! Tried some single files, some mixed archives/normal files and didn't encounter any problems.

brp-david:
Getting an error when scanning a deep folder structure, this is present in the commit before you merged my changes aswell (but not in the master branch). I haven't debugged it more than this but I'll do that later

PR #592, opened 2016-03-07
Add ItaSA Subtitle Provider

Elettronik:
Added italiansubs.net provider, based on the xml apis of site.


Diaoul:
If you work only with XML don't use BeautifulSoup, use [etree](https://github.com/Diaoul/subliminal/blob/develop/subliminal/providers/podnapisi.py#L8-L14)

subliminal/providers/itasa.py
```diff
@@ -0,0 +1,390 @@
...
+
+
+class ItaSAProvider(Provider):
+    languages = {Language('ita')}
+
+    video_types = (Episode,)
+
+    server_url = 'https://api.italiansubs.net/api/rest/'
+
+    apikey = 'd86ad6ec041b334fac1e512174ee04d5'
```
Diaoul: Where does this comes from?

subliminal/providers/itasa.py
```diff
@@ -0,0 +1,390 @@
...
+                    return show_id
+
+            next = root.find('next')
+
+        # No matches found
+        logger.warning('Show id not found: suggestions does not match')
+
+        return None
+
+    def get_show_id(self, series, country_code=None):
```
Diaoul: Is this relevant for ItaSA? Searching with country first?

subliminal/providers/itasa.py
```diff
@@ -0,0 +1,390 @@
...
+            with ZipFile(io.BytesIO(content)) as zf:
+                if len(zf.namelist()) > 1:
+
+                    for name in enumerate(zf.namelist()):
+
+                        if name[0] == 0:
+                            # First elemnent
+                            sub.content = fix_line_ending(zf.read(name[1]))
+                            sub.full_data = name[1]
+                        else:
```
Diaoul: Why treat additional subtitles separately?

subliminal/providers/itasa.py
```diff
@@ -0,0 +1,390 @@
...
+            show_id = self._search_show_id(series)
+
+        return show_id
+
+    def _download_zip(self, sub_id):
+        # download the subtitle
+        logger.info('Downloading subtitle %r', sub_id)
+
+        # log into main site
+        if not self.login_itasa:
```
Diaoul: Download isn't the right place for login.

subliminal/providers/itasa.py
```diff
@@ -0,0 +1,390 @@
...
+        return self._format
+
+    @format.setter
+    def format(self, value):
+        if not value:
+            self._format = 'HDTV'
+        elif value and value.lower() == 'normale':
+            self._format = 'HDTV'
+        else:
+            self._format = value
```
Diaoul: Please handle conversions separately.

subliminal/providers/itasa.py
Elettronik: What do you mean?

subliminal/providers/itasa.py
```diff
@@ -0,0 +1,390 @@
...
+                        season,
+                        episode,
+                        subtitles.version.text,
+                        subtitle.find('name').text)
+
+                    subtitles.append(sub)
+
+            next = root.find('next')
+
+        # Dowload the subs found, can be more than one in zip
```
Diaoul: Why download the subtitles? This should be done in the `download` method instead.

setup.py
```diff
@@ -66,7 +66,8 @@
               'podnapisi = subliminal.providers.podnapisi:PodnapisiProvider',
               'subscenter = subliminal.providers.subscenter:SubsCenterProvider',
               'thesubdb = subliminal.providers.thesubdb:TheSubDBProvider',
-              'tvsubtitles = subliminal.providers.tvsubtitles:TVsubtitlesProvider'
+              'tvsubtitles = subliminal.providers.tvsubtitles:TVsubtitlesProvider',
+              'itasa = subliminal.providers.itasa:ItaSAProvider'
```
Diaoul: Please keep this alphabetically sorted

subliminal/providers/itasa.py
```diff
@@ -0,0 +1,390 @@
...
+from .. cache import SHOW_EXPIRATION_TIME, region
+
+from .. exceptions import AuthenticationError, ConfigurationError, TooManyRequests
+
+from .. subtitle import (Subtitle, fix_line_ending, guess_matches, guess_properties, sanitize_string,
+                         sanitized_string_equal)
+
+from .. video import Episode
+
+logger = logging.getLogger(__name__)
```
Diaoul: Is this pep8 valid? Please keep import syntax consistent with other providers.

subliminal/providers/itasa.py
```diff
@@ -0,0 +1,390 @@
...
+
+    itasa_formats = {'1080i': '1080i',
+                     '1080p': '1080p',
+                     '720p': '720p',
+                     'bdrip': 'bdrip',
+                     'bluray': 'bluray',
+                     'dvdrip': 'dvdrip',
+                     'hdtv': 'normale',
+                     'hr': 'hr',
+                     'web-dl': 'web-dl'}
```
Diaoul: No need to specify them all, just the exceptions. BTW, why the conversion anyway?

Diaoul:
Thanks for the PR, I left a few comments.

subliminal/providers/itasa.py
Diaoul: How did you get it

subliminal/providers/itasa.py
Elettronik: ItaSA list on site one subtitle, but in the resultant zip, could be more than one. 
Can be subtitle version 2, with v2 suffix or repack with R suffix.

subliminal/providers/itasa.py
Elettronik: This is a particular to itasa. the site login is needed to download subtitles (give session id).
Is it better to do both login in initialize?

subliminal/providers/itasa.py
Elettronik: Where is better? In the part I create the subtitle instance?

subliminal/providers/itasa.py
Elettronik: Which line do you mean? The validation with PEP8 is ok. if you can tell me the line I will copy style from addic7ed.

subliminal/providers/itasa.py
Elettronik: I want to make a generic mechanism, but maybe is better to convert on the fly in the code the create the subtitle instance

subliminal/providers/itasa.py
Diaoul: Yes, basically this class should be a dumb container.

subliminal/providers/itasa.py
Diaoul: * Group from/imports by "kind" (native modules imports, dependencies imports, relative imports) and sort them alphabetically with one new line between each group.
* Sort imports alphabetically (first constants, then classes, then functions).

See other modules for example. This is not critical but just to keep the style consistent.

subliminal/providers/itasa.py
Diaoul: You can do `itasa_formats.get(original_format, original_format)`, it will then fall back on `original_format` by default.

subliminal/providers/itasa.py
Diaoul: You mean there are two kind of logins?

subliminal/providers/itasa.py
Diaoul: Could it be a subtitle for a completely different release?

subliminal/providers/itasa.py
Elettronik: The site is a community of italian subber. I requested a developer access and they give me the apikey,  and the api documentation. The only restriction they put on the api usage is that application/library should be public domain.
For more info the site www.italiansubs.net is their homepage.

subliminal/providers/itasa.py
Elettronik: As I saw on addic7ed provider, from which i copied, same show can have local variation such  UK, US. 
This fit that cases.

subliminal/providers/itasa.py
Elettronik: No, it is realative to same release, just a different version. Sometime is a better resync or commented by translator.

subliminal/providers/itasa.py
Elettronik: The fact is that if you want to download you need a session cookie from main site login, which cannot be retrieved by the api login. Maybe is better to do both login in initialize method?

Elettronik:
Done requested changes

subliminal/providers/itasa.py
Diaoul: OK, that's weird. Anyways, handle both logins in initialize and assume you are logged in both list and download methods.

Elettronik:
can this change be included in a minor like 1.2, so i can integrate in other project?
Btw I just fixed the login in initialization

Diaoul:
It will be in next major along with other providers (shooter and legendastv). 2.0 is ready, I just need to review and merge those new providers.

Diaoul:
Please rebase on develop and squash commits, I will merge and finalize this locally after a more in-depth analysis.

Elettronik:
Hope I did ok, I'm fairly new to git collaboration.
I rebased on origin/develop and squashed all commits, then forced push.

Elettronik:
I fixed a few api change from the develop branch, but I have a doubt about how to treat dots:
```
def test_get_show_ids_quote():
    with ItaSAProvider() as provider:
        show_ids = provider._get_show_ids()
    assert 'marvels agents of s h i e l d' in show_ids
    assert show_ids['marvels agents of s h i e l d'] == 4430
```
In this test whentI fetch the show the result i "Marvel's agent of S.H.I.E.L.D", but in other addic7ed provider is tested against "marvels agents of shield". Is it correct to remove all dots in the string like the previous version of sanitize_string?

Issue #536, opened 2015-11-24
Subtitles fron NapiProjekt are not in .srt format


ddanielczyk:
Hi,

So all subtitles downloaded from napiprojekt, are in mDVD format, not srt. I noticed though that Subliminal saves them with .srt extension, but if player supports only .srt, they will not work.

Is there a chance to use eg. subconv to convert such subtitles from mDVD to srt format on fly ?

Thanks
##

Diaoul:
I believe this has been discussed in another issue but I can't find it. This would required proper encoding detection with #528 and subtitle parsing with #462.

bogdal:
Here is the [link to the discussion](https://github.com/Diaoul/subliminal/pull/489#issuecomment-152475830).

Diaoul:
Thanks @bogdal. @DDamian can you provide more details https://github.com/Diaoul/subliminal/blob/develop/CONTRIBUTING.md ?

DDamian:
Hi, sure, just tell me what details you need. I am using it from SickRage, but anyway, it doesn't matter, because all subtitles from NapiProjekt are always in mDVD format and as far as I know, they're always encoded Windows-1250. So Subliminal and SickRage(using subliminal) will behave the same way here.

I'm not sure if we could maybe try assume that if subtitles downloader is NapiProjekt, then force Windows-1250 -> UTF8 encoding - this would solve us the encoding problem. (But this is just an idea).

About converting subtitles from mDVD to srt, it should be quite simple, we just need to get framerate of video and then it's quite easy to convert. As far as I know Subconv does all that, but I guess it could be easily done manually, it's up to you. I'm willing to help.

Thanks

DDamian:
BTW, why would it require proper encoding detection? We can convert mDVD->srt in any encoding and just save it with this encoding to .srt. Can't we ?

DDamian:
One more, just tested with latest dev from Subliminal. When it downloads from napiprojekt, subtitles are Windows-1250, but if I set Subliminal to convert to utf8 - it guess original encoding to be Iso-8859-2 and then convert to utf8. Its successfull, but FEW polish characters are displayed incorrectly, not all of them, just few. And this FEW is because of difference between Windows-1250 and Iso-8859-2.  So once again I think we could succcessfully convert napiprojekt subs to UTF-8, but we must tell subliminal, that original encoding is Windows-1250, not ISO-8859-2.

Diaoul:
[...]

PR #605, opened 2016-04-02
Add itasa provider + fixes

miigotu:
If you want to merge this that would be great, otherwise it is mostly to keep track of the patches we have in sickrage (I will try and keep this updated with your develop branch).

Now only adds itasa, in an unregistered state.

### DIFF
--- a/subliminal/cli.py
+++ b/subliminal/cli.py
@@ -215,6 +215,7 @@ config_file = 'config.ini'
 @click.group(context_settings={'max_content_width': 100}, epilog='Suggestions and bug reports are greatly appreciated: '
              'https://github.com/Diaoul/subliminal/')
 @click.option('--addic7ed', type=click.STRING, nargs=2, metavar='USERNAME PASSWORD', help='Addic7ed configuration.')
+@click.option('--itasa', type=click.STRING, nargs=2, metavar='USERNAME PASSWORD', help='ItaSA configuration.')
 @click.option('--legendastv', type=click.STRING, nargs=2, metavar='USERNAME PASSWORD', help='LegendasTV configuration.')
 @click.option('--opensubtitles', type=click.STRING, nargs=2, metavar='USERNAME PASSWORD',
               help='OpenSubtitles configuration.')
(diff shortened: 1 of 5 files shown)


miigotu:
not sure how to add cassettes, or fixup the opensubtitles cassettes.

If you can take care of that I can/will fix the other tests if needed.

Diaoul:
Itasa isn't merged yet and code not reviewed. That means a lot of things can change like for legendas.
#536 is the reason why NapiProjekt is disabled.
Why add self.session to `__init__`? Why add "missing" ones? I have a warning on my IDE but nor pep8 neither pyflakes complain about this so why add code if it doesn't do anything? Sometimes one ought to be clever than best practices. Not sure it applies here anyway because documentation clearly states `initialize` is to be called if not using the `with` statement. I agree this is not ideal but I don't like the solution I'm left with. The alternative is to remove `initialize` and use `__init__` for that purpose, might be a good solution but I need to review consequences. PR accepted if you're willing to contribute.
You cannot catch everything that can go wrong in Python, don't bother with ConfigurationError on empty string. The idea here is to catch missing parameters that come together. If you filled the parameter with crap you cannot expect anything good from this.
Agreed for the raw string prefix in regex.

miigotu:
Don't merge yet, I will make some adjustments, like not enabling napiprojekt and split out some of the other changes into smaller PR.

My IDE (editor really) shows pep8, pyflakes, and pylint output.

http://pylint-messages.wikidot.com/messages:w0232
http://pylint-messages.wikidot.com/messages:w0201

About the username/passwords:
Checking only against `None` is fine if it is only being used via cli, as the arguments cannot be empty strings easily. But when using it as a module I would have to explicitly do `sickbeard.BLAH_PASSWORD or None` to be sure it is handled correctly in the module.

While that is ok and easy for me to do, if someone does not know of this quirk and is implementing subliminal as a module into their project it can cause them headaches figuring out why it is trying to log in when they didn't provide a username or password and subsequently becoming a disabled_provider in the pool. It just seems to have no negative effect and makes it a little easier on 3rd parties, because an empty username or password can never log in to any site anyways.

I will contribute as much back as I can, because I think its silly to keep patching downstream and it's sort of part of the deal to help back when you use someone else's work.

miigotu:
@Diaoul I split out the straight forward stuff into #617

miigotu:
https://github.com/SickRage/SickRage/issues/1457

fernandog:
A newer version is available in PR 804 from @Elettronik

PR #610, opened 2016-04-02
Improving season detection and handling missing season info

ratoaq2:
- Handling titles where there's no season information or the season description is in english
- Added regression tests

I'm creating several small PR in order to facilitate review/accept/reject. I'm also not sure if you've already fixed any or all of these issues.

### DIFF
--- a/subliminal/providers/legendastv.py
+++ b/subliminal/providers/legendastv.py
@@ -33,7 +33,7 @@ rarfile.PATH_SEP = '/'
 type_map = {'M': 'movie', 'S': 'episode', 'C': 'episode'}
 
 #: BR title season parsing regex
-season_re = re.compile(r' - (?P<season>\d+)(\xaa|a) temporada', re.IGNORECASE)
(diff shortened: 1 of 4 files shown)


fernandog:
@Diaoul tested this in SickRage and its working

Diaoul:
@fernandog: I can see from the unittests by myself

Please rebase and remove the second commit, I prefer a `in` check [here](https://github.com/Diaoul/subliminal/blob/develop/subliminal/providers/legendastv.py#L334) so we won't match on `None` however improbable that is and we keep `search_titles` consistent, i.e. no information == no key in the dict.

subliminal/providers/legendastv.py
```diff
@@ -229,7 +229,8 @@ def search_titles(self, title):
                     title['season'] = int(source['temporada'])
                 else:
                     match = season_re.search(source['dsc_nome_br'])
-                    title['season'] = int(match.group('season'))
+                    if match:
+                        title['season'] = int(match.group('season'))
```
Diaoul: Please add a warning if the match failed as I believe this is quite unexpected, right?

PR #612, opened 2016-04-02
Changing cache configuration for legendastv provider

ratoaq2:
- Removing caching when listing archives
> when an episode or a movie is release, you'll have a growing list of archives for the next couple of days. If we cache this URL (carrega_legendas_busca_filme) we might wait for a long time to get a subtitle that was added afer your initial search.

- Changing release names cache to 15 min
> It's a common practice that the uploaders re-upload the archivers adding more releases to it. And that happens in a high frequency for the first couple of days after the release.

### DIFF
--- a/subliminal/providers/legendastv.py
+++ b/subliminal/providers/legendastv.py
@@ -42,7 +42,7 @@ downloads_re = re.compile(r'(?P<downloads>\d+) downloads')
 rating_re = re.compile(r'nota (?P<rating>\d+)')
 
 #: Timestamp parsing regex
-timestamp_re = re.compile(r'(?P<day>\d+)/(?P<month>\d+)/(?P<year>\d+) - (?P<minute>\d+):(?P<second>\d+)')
+timestamp_re = re.compile(r'(?P<day>\d+)/(?P<month>\d+)/(?P<year>\d+) - (?P<hour>\d+):(?P<minute>\d+)')
 
 #: Cache key for releases
 releases_key = __name__ + ':releases|{archive_id}'
@@ -251,7 +251,7 @@ class LegendasTVProvider(Provider):
 
         return titles
 
-    @region.cache_on_arguments(expiration_time=timedelta(minutes=15))
+    @region.cache_on_arguments(expiration_time=timedelta(minutes=15).total_seconds())
     def get_archives(self, title_id, language_code):
         """Get the archive list from a given `title_id` and `language_code`.
 


Diaoul:
I need to compare with other providers to ensure consistency of how to treat episode related information but I tend to agree with you.

However the release name cache should take into account the upload timestamp and use that to determine when to invalidate the cache.

ratoaq2:
Please, provide your feedback.

I'm using the timestamp to calculate the expiration_time to be used.
If the archive timestamp is too recent, the expiration_time will be short, e.g: archive was updated 2h ago, releases cache will expire in 15 min. I set the maximum expiration_time to be 1 day.

subliminal/providers/legendastv.py
```diff
@@ -266,12 +275,19 @@ def get_archives(self, title_id, language_code):
...
-                match = rating_re.match(archive_soup.find('p', class_='data').text)
+                match = rating_re.match(data_text)
                 archive.downloads = int(match.group('downloads'))
                 if match.group('rating') is not None:
                     archive.rating = int(match.group('rating'))
 
+                # timestamp
+                match = timestamp_re.search(data_text)
+                if match:
+                    archive.timestamp = datetime.datetime.strptime(match.group('timestamp'), '%d/%m/%Y - %H:%M')
```
Diaoul: Why not name groups and use datetime constructor? `datetime(**match.groups)`

ratoaq2: Ok... I'll simplify the expiration_time as well:
- archivers older than 1 week = 1 day expiration time
- archivers newer than 1 week = 1 hour expiration time

Sounds good?

Diaoul: I will change that part to something else. The cache should store the tuple `(timestamp, releases)` so that when we get from the cache we can compare timestamp and decide to use it or not.

Diaoul:
I think archives should be cached for at least a few minutes. Say you search a whole season folder of a particular show (common use case). You will end up doing as many requests to `get_archives` as your number of files in the season folder.

subliminal/providers/legendastv.py
Diaoul: Nevermind, your solution is better. Didn't read the `region.get` documentation properly :+1:

ratoaq2:
Now I got confused. Let me summarize what I understood.
- I shouldn't change to cache `(timestamp, releases)`
- I should change to `archivers older than 1 week` / `archivers newer than 1 week`
- I should choose lower `expiration_time` other than 1h and 1day

Diaoul:
Sorry about that, I'm currently working on it too. Let me make a commit and comment on it.

subliminal/providers/legendastv.py
```diff
@@ -266,12 +275,19 @@ def get_archives(self, title_id, language_code):
...
                 # match rating
-                match = rating_re.match(archive_soup.find('p', class_='data').text)
+                match = rating_re.match(data_text)
                 archive.downloads = int(match.group('downloads'))
                 if match.group('rating') is not None:
                     archive.rating = int(match.group('rating'))
 
+                # timestamp
+                match = timestamp_re.search(data_text)
+                if match:
```
Diaoul: Do you have an example where it doesn't match?

ratoaq2: I don't remember the case right now, but I do remember that exists entries without it (I believe that they are related to old entries that came from a site migration).

And I believe that some timestamps might not represent a valid date and time (e.g.: 13/13/2010 - 11:11), but since I'm not sure about this last case, better to not handle it and if we get a case, then solve it.

ratoaq2: I couldn't find a single case right now that it doesn't match. Up to you to decide whether to handle non matches or not. If someone reports a bug about it, it would be an easy fix anyway.

Diaoul: Do you know the locale of the timestamp? If we don't parse the date right we might miss some updated archives.

ratoaq2: Location is Brazil - São Paulo: GMT - 3

Note: during summer, the summer time (end of October to end of February): GMT - 2

subliminal/providers/legendastv.py
```diff
@@ -359,8 +375,16 @@ def query(self, language, title, season=None, episode=None, year=None):
...
+                if a.timestamp:
+                    delta = max(0, (datetime.datetime.now() - a.timestamp).total_seconds())
+                    # recently uploaded archives have short expiration time
+                    expiration_time = min(RELEASES_EXPIRATION_TIME, delta / 8)
+                else:
+                    expiration_time = RELEASES_EXPIRATION_TIME
+
                 # attempt to get the releases from the cache
-                releases = region.get(releases_key.format(archive_id=a.id))
+                releases = region.get(releases_key.format(archive_id=a.id), expiration_time=expiration_time)
```
Diaoul: What about simply `expiration_time = max(0, (datetime.now() - a.timestamp).total_seconds())`? This will use the cache whenever possible.

Diaoul: I think the max isn't necessary unless we have an upload time in the future which. We also might want some margin of error.
Issues with this will be hard to identify so we better think this through now :p

ratoaq2: If we have `a.timestamp = 01.01.2010` then the expiration time will be > 6 years. That's too much.

subliminal/providers/legendastv.py
Diaoul: So [this](https://en.wikipedia.org/wiki/America/Sao_Paulo) is the one

subliminal/providers/legendastv.py
Diaoul: Well if it's not in the cache it will be cached and if we want to get it another time it will get it from cache.
What's the issue with getting the issue from the cache if we know it hasn't changed?

subliminal/providers/legendastv.py
ratoaq2: Exactly

subliminal/providers/legendastv.py
ratoaq2: The only issue would be if you would like to manage the cache size or limit it to not go beyond certain value. I know that cli stores the cache on the disk. The python api leaves the decision to the client/developer whether to use disk or memory cache.

Diaoul: This is a common misunderstanding. Cache is ever-growing. Nothing is ever deleted from cache. Deleting expired keys isn't part of dogpile.cache's API and never will be. This is why I don't to store binary or whole HTTP requests in cache. With the amount of requests done by subliminal that would mean a really fast paced growing cache.

Only some backends do cache clean up of expired keys based on the backend expiration time which is different from dogpile.cache's.
Documentation says backend's expiration time should be greater than dogpile.cache's but I think a lesser value would trigger a cache refresh, if that's what you're looking for.

Diaoul:
Let me know what you think b9c4c86e1fefe4945f1bbce5e9a35b032b6c016c

subliminal/providers/legendastv.py
ratoaq2: This is something that I felt it was missing from the dogpile.cache's API: eviction policy. My background is related to Java and I'm quite used to its cache API, using LRU eviction policy among other things.

ratoaq2:
I'll test it now

Shouldn't [this](https://github.com/Diaoul/subliminal/commit/b9c4c86e1fefe4945f1bbce5e9a35b032b6c016c#diff-e33b1676e0aad8bc5e406d5983fe5bc7R12) be added as a dependency

ratoaq2:
A small issue:

the cache configuration [here](https://github.com/Diaoul/subliminal/commit/b9c4c86e1fefe4945f1bbce5e9a35b032b6c016c#diff-e33b1676e0aad8bc5e406d5983fe5bc7R254) is missing `total_seconds()`:
`@region.cache_on_arguments(expiration_time=timedelta(minutes=15).total_seconds())`.

The second time, we get an exception:
```
ERROR:subliminal.core:Unexpected error in provider 'legendastv'
  File "/home/user/workspace/git/subliminal/subliminal/core.py", line 118, in list_subtitles_provider
    return self[provider].list_subtitles(video, provider_languages)
  File "/home/user/workspace/git/subliminal/subliminal/providers/legendastv.py", line 437, in list_subtitles
    return [s for l in languages for s in self.query(l, title, season=season, episode=episode, year=video.year)]
  File "/home/user/workspace/git/subliminal/subliminal/providers/legendastv.py", line 371, in query
    for a in self.get_archives(title_id, language.legendastv):
  File "/home/user/subliminal/local/lib/python2.7/site-packages/dogpile.cache-0.5.7-py2.7.egg/dogpile/cache/region.py", line 1053, in decorate
    should_cache_fn)
  File "/home/user/subliminal/local/lib/python2.7/site-packages/dogpile.cache-0.5.7-py2.7.egg/dogpile/cache/region.py", line 657, in get_or_create
    async_creator) as value:
  File "/home/user/subliminal/lib/python2.7/site-packages/dogpile.core-0.4.1-py2.7.egg/dogpile/core/dogpile.py", line 158, in __enter__
    return self._enter()
  File "/home/user/subliminal/lib/python2.7/site-packages/dogpile.core-0.4.1-py2.7.egg/dogpile/core/dogpile.py", line 98, in _enter
    generated = self._enter_create(createdtime)
  File "/home/user/subliminal/lib/python2.7/site-packages/dogpile.core-0.4.1-py2.7.egg/dogpile/core/dogpile.py", line 116, in _enter_create
    if not self._is_expired(createdtime):
  File "/home/user/subliminal/lib/python2.7/site-packages/dogpile.core-0.4.1-py2.7.egg/dogpile/core/dogpile.py", line 79, in _is_expired
    time.time() - createdtime > self.expiretime
TypeError: can't compare datetime.timedelta to float
```

ratoaq2:
Another issue that I've found is:

- First run, releases are stored in the cache. The archive timestamp was from yesterday, let's say 1 day, so the expiration_time used was 24h
- Second run, I hardcoded expiration_time = 60 (seconds). Since I did run the second time right after the first run, it hit the cache instead of discarding it.

I think it's not working as we were expecting. Probably you need to store the tuple `(timestamp, releases)` in order to discard the value if the timestamp doesn't match.

ratoaq2:
Now I have short time, I did a quick/dirty fix to discard when timestamp doesn't match.

ratoaq2:
I'll need time to test it. I don't know if I provided a real case. I'll do it later

ratoaq2:
I performed more tests and it works like expected :+1: 

So, I updated this PR with 2 minor fixes: Missing `total_seconds()` and fixing the timestamp regex, since the time part should be `hour`:`minute` and not `minute`:`second`.
A 3rd issue: the missing pytz dependency is not part of this PR. I believe that you will address anyway.

Diaoul:
All done with latest develop. Thanks for your contributions.

Issue #603, opened 2016-04-01
TVDB refiner pick wrong series


ratoaq2:
For the latest develop version (2.0-rc1)
For CSI tvdb refiner pick the wrong series.

```
subliminal.py --debug --legendastv pass user download -f -l pt-BR -p legendastv -m 97 CSI.S15E18.720p.HDTV.X264.DIMENSION.mkv
Collecting videos  [####################################]  100%  CSI.S15E18.720p.HDTV.X264.DIMENSION.mkvDEBUG:subliminal.cli:Collecting path CSI.S15E18.720p.HDTV.X264.DIMENSION.mkv
INFO:subliminal.core:Refining video with metadata
INFO:subliminal.core:Refining video with tvdb
INFO:subliminal.refiners.tvdb:Searching series 'CSI'
DEBUG:subliminal.refiners.tvdb:Found 9 results
DEBUG:subliminal.refiners.tvdb:Found result for original series without year
DEBUG:subliminal.refiners.tvdb:Found series {u'seriesId': u'', u'status': u'Continuing', u'added': u'2011-10-30 01:27:22', u'airsTime': u'20:00', u'network': u'TV Nova', u'rating': u'', u'overview': u'Crimi series inspired by real cases.', u'lastUpdated': 1440591672, u'zap2itId': u'', u'seriesName': u'CSI Andel', u'siteRating': 6, u'firstAired': u'2008-09-01', u'addedBy': 8907, u'networkId': u'', u'genre': [u'Drama'], u'airsDayOfWeek': u'Monday', u'runtime': u'55', u'banner': u'graphical/253208-g.jpg', u'id': 253208, u'imdbId': u'tt1322074', u'aliases': []}
INFO:subliminal.refiners.tvdb:Getting series episode 15x18
WARNING:subliminal.refiners.tvdb:No results for episode
```

Later in the logs:
```
INFO:subliminal.providers.legendastv:Searching title u'csi andel'
```

Diaoul:
What 's the good one on tvdb?

ratoaq2:
id = 72546

Diaoul:
From the filename it's clear subliminal cannot find a clear match on tvdb for this. I think I can reject candidates based on a maximal Levenshtein distance which could be something like `max(int(0.2*len(title)), 2)`
So the match title cannot differ too much from the original title.
Levenshtein distance might not be best suited for that as I don't want words to differ, CSI cannot be CSA.
[...]

Issue #602, opened 2016-04-01
TVDB refiner: ValueError: time data '' does not match format '%Y-%m-%d'


ratoaq2:
For the latest develop version (2.0-rc1)
`subliminal --debug --legendastv user pass download -f -l pt-BR -p legendastv -m 97 The.100.S03E09.720p.HDTV.x264-AVS.mkv`

```
Collecting videos  [####################################]  100%  The.100.S03E09.720p.HDTV.x264-AVS.mkvDEBUG:subliminal.cli:Collecting path The.100.S03E09.720p.HDTV.x264-AVS.mkv
INFO:subliminal.core:Refining video with metadata
INFO:subliminal.core:Refining video with tvdb
INFO:subliminal.refiners.tvdb:Searching series 'The 100'
DEBUG:subliminal.refiners.tvdb:Found 45 results
DEBUG:subliminal.refiners.tvdb:Found result for original series without year
DEBUG:subliminal.refiners.tvdb:Found series {u'seriesId': u'', u'status': u'Continuing', u'added': u'2013-08-10 11:43:14', u'airsTime': u'', u'network': u'', u'rating': u'', u'overview': u'Planet Earth is stacked with amazing feats of engineering and construction, a dazzling array of machines and vehicles, an awe inspiring great outdoors and a near limitless number super skilled, super humans\r\nWhat we want to know exactly how much of all this we\u2019ve got, where it all is and what it exactly it does\u2026\r\nIn each episode we will carry out the ultimate, exhaustive stock take of that particular category (from skyscrapers to mountains, boats to special forces, railways to fire-fighters\u2026discovering exactly where every single example is located and in doing so answer those questions we have about the world that have never been answered until now.\r\nA fast paced entertaining mix of graphics, archive and location reports together will be the ultimate information countdown.', u'lastUpdated': 1385372430, u'id': 272209, u'seriesName': u'100%', u'firstAired': u'', u'addedBy': 391110, u'banner': u'graphical/272209-g.jpg', u'networkId': u'', u'genre': [u'Documentary'], u'airsDayOfWeek': u'', u'runtime': u'1', u'siteRating': 0, u'zap2itId': u'', u'imdbId': u'', u'aliases': []}
ERROR:subliminal.core:Failed to refine video
Traceback (most recent call last):
  File "/home/user/workspace/git/subliminal/subliminal/core.py", line 541, in refine
    refiner_manager[refiner].plugin(video, **kwargs)
[...]

PR #614, opened 2016-04-02
Fix for #602 and #603

ratoaq2:
- Fixing date parse when there's no firstAired entry. (Fix for #602)
- Use case insensitive (e.g.: When searching for 'The 100' will not choose '100%', but the correct 'The 100' result). (Related to #602 and #603)
- 'aliases' for better series matching (e.g: CSI will be chosen since CSI is in the aliases list of CSI: Crime Scene Investigation) (Fix for #603)

### DIFF
--- a/subliminal/refiners/tvdb.py
+++ b/subliminal/refiners/tvdb.py
@@ -209,7 +209,7 @@ def search_series(name):
         key = 0
         if series['status'] != 'Continuing':
             key += 1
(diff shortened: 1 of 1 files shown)


ratoaq2:
This was fixed by https://github.com/Diaoul/subliminal/commit/7c2023b4b308f78d3907e9f10b11f96289dea56c

PR #617, opened 2016-04-03
Small adjustments, split out from #605

miigotu:
Use raw strings for regex literals
Tweak username and password conditionals in providers to also catch empty strings as invalid
Add some missing `__init__` methods, and add class properties that were defined outside of `__init__`

### DIFF
--- a/subliminal/providers/addic7ed.py
+++ b/subliminal/providers/addic7ed.py
@@ -86,19 +86,20 @@ class Addic7edProvider(Provider):
     server_url = 'http://www.addic7ed.com/'
 
     def __init__(self, username=None, password=None):
-        if username is not None and password is None or username is None and password is not None:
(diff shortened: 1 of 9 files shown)


Diaoul:
Raw strings with 25ece03102cd43581c972cc0bd99d4bd47140392

miigotu:
I still think the other tweaks are beneficial, but you're the boss =P

Diaoul:
That's why I didn't close :)

miigotu:
bwaha, ok. Ill fix the conflicts then xD

fernandog:
Merged in https://github.com/Diaoul/subliminal/commit/2d266cdae93129d7c7bd5ffe73e25fac2200cc94#diff-c9c71e810cede81785a6732f7362417a

PR #629, opened 2016-04-10
TVDB refiner should discard result if year doesn't match

ratoaq2:
- A small fix to discard the result if the year doesn't match
- Missing first argument in logging call.


Diaoul:
I have a pending fix locally

Diaoul:
See d5b4a4f226ef9fc28dcb14f8ac2b515e39d92414

PR #631, opened 2016-04-10
TVDB refiner: failing tests for series name with year

ratoaq2:
This is only the failing test cases for series with year in their name, e.g:
- Once Upon a Time (2011)
- House of Cards (2013)

No cassettes since I don't have a tvdb API key.

I hope this could be useful.


Diaoul:
I have a pending fix locally

Diaoul:
See d5b4a4f226ef9fc28dcb14f8ac2b515e39d92414

PR #651, opened 2016-04-25
Better check of username/password for ltv

fernandog:
When not username and not password, it didn't raise ConfigurationError
~~"if year is not None" is the same as "if year"~~ 

@ratoaq2

### DIFF
--- a/subliminal/providers/legendastv.py
+++ b/subliminal/providers/legendastv.py
@@ -122,7 +122,7 @@ class LegendasTVSubtitle(Subtitle):
                 matches.add('series')
 
             # year (year is based on season air date hence the adjustment)
-            if video.original_series and self.year is None or video.year and video.year == self.year - self.season + 1:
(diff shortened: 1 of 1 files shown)


Diaoul:
This is pure stylistic changes without unittests to back it up. `None` test is more accurate than testing as boolean.

fernandog:
what about all[] that checks for 4 combinations of user/password instead of the existing 2?

Diaoul:
The goal is to check that a parameter was actually passed even if using an optional argument, checking validity of username password is out of scope of subliminal.
If on the server the password should be 6 characters long, should subliminal check that too? I say no.

Actually I was thinking making those arguments positional but then the way the constuctor is called using provider configs should be changed too.
Maybe 2.0 is a good milestone for that change, if this introduces a breacking change.

fernandog:
but for ltv to work it need username AND password. Only username, it will fail. Only password it will fail.

Issue #660, opened 2016-05-16
KeyError: 'year' (legendastv provider)


ratoaq2:
Latest develop version
```
commit 14707b68ca1f8ed6cdd78697c6bf6055a861ca2d
Author: Antoine Bertin <diaoulael@gmail.com>
Date:   Sun May 1 18:44:54 2016 +0200

    Ensure scan_videos continues if unrar is not installed
```

When searching movies subtitles using legendastv provider:
`subliminal --debug --legendastv user pass download -f -l pt-BR -p legendastv -m 97 "The.Boy.Next.Door.2015.1080p.BluRay.x264-SPARKS.mkv"`

The following exception happens:
```
Collecting videos  [####################################]  100%  The.Boy.Next.Door.2015.1080p.BluRay.x264-SPARKS.mkvDEBUG:subliminal.cli:Collecting path The.Boy.Next.Door.2015.1080p.BluRay.x264-SPARKS.mkv
INFO:subliminal.core:Refining video with metadata
INFO:subliminal.core:Refining video with omdb
DEBUG:subliminal.refiners.omdb:Found 1 results
DEBUG:subliminal.refiners.omdb:Found result with matching year
DEBUG:subliminal.refiners.omdb:Found movie {u'imdbID': u'tt3181822', u'Year': u'2015', u'Type': u'movie', u'Poster': u'http://ia.media-imdb.com/images/M/MV5BMTgxNTEyMTYzOV5BMl5BanBnXkFtZTgwNzQ4OTg5MjE@._V1_SX300.jpg', u'Title': u'The Boy Next Door'}
Collecting videos  [####################################]  100%                                                     
1 video collected / 0 video ignored / 0 error
Downloading subtitles  [####################################]  100%  The.Boy.Next.Door.2015.1080p.BluRay.x264-SPARKS.mkvINFO:subliminal.core:Listing subtitles with provider 'legendastv' and languages set([<Language [pt-BR]>])
INFO:subliminal.core:Initializing provider legendastv
INFO:subliminal.providers.legendastv:Logging in
DEBUG:subliminal.providers.legendastv:Logged in
INFO:subliminal.providers.legendastv:Searching title u'the boy next door'
DEBUG:subliminal.providers.legendastv:Found 3 titles
ERROR:subliminal.core:Unexpected error in provider 'legendastv'
Traceback (most recent call last):
  File "build/bdist.linux-x86_64/egg/subliminal/core.py", line 118, in list_subtitles_provider
[...]

PR #661, opened 2016-05-16
Fix for #660

ratoaq2:
`year` is not always available for movie titles


ratoaq2:
Fixed by https://github.com/Diaoul/subliminal/commit/e2331bef2c827406f92b1696b96241137163fe4f

PR #680, opened 2016-08-11
Fixed subscenter provider to support new URLs.

ofir123:
Subscenter provider stopped working because site URLs have changed.
I've fixed the provider to work with the new URLs and to handle redirections within the site.


Diaoul:
Thanks for the fix.
Can you squash and rebase against master as this will be released as a hotfix?

ofir123:
Sure! Opened a separate one for the master.

PR #686, opened 2016-09-04
Change podnapisi URL to .eu

fernandog:
PR in behalf of @medariox

 The issue is that .net redirects to https and the default setting for allow_redirects (requests) is False for podnapisi. So we have to avoid the redirect by using https with .net or http with .eu, or set allow_redirects=True in podnapisi provider.

### DIFF
--- a/subliminal/providers/podnapisi.py
+++ b/subliminal/providers/podnapisi.py
@@ -82,7 +82,7 @@ class PodnapisiProvider(Provider):
     """Podnapisi Provider."""
     languages = ({Language('por', 'BR'), Language('srp', script='Latn')} |
                  {Language.fromalpha2(l) for l in language_converters['alpha2'].codes})
-    server_url = 'http://podnapisi.eu/subtitles/'
+    server_url = 'https://podnapisi.net/subtitles/'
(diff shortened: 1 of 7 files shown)


fernandog:
@Diaoul why travis failed? Can't understand why

fernandog:
@Diaoul can you comment please?

Diaoul:
Let's use HTTPS instead.

fernandog:
@Diaoul done

fernandog:
@Diaoul can you check this please?

subliminal/providers/podnapisi.py
```diff
@@ -82,7 +82,7 @@ class PodnapisiProvider(Provider):
     """Podnapisi Provider."""
     languages = ({Language('por', 'BR'), Language('srp', script='Latn')} |
                  {Language.fromalpha2(l) for l in language_converters['alpha2'].codes})
-    server_url = 'http://podnapisi.net/subtitles/'
+    server_url = 'http://podnapisi.eu/subtitles/'
```
Diaoul: Shoudln't that be `https://podnapisi.net/subtitles/`?

PR #696, opened 2016-09-30
OpenSubtitles: match by 'tag' only if same imdb or when video has no imdb id info

fernandog:
@Diaoul 
This fixes when searching by `The Fall S03E01` and I get results from  `doc.martin.s03e01.(24 september 2007).[tvrip (xvid)]-spa.srt` 

It's a `tag` match. But IMDB is different.


subliminal/providers/opensubtitles.py
```diff
@@ -58,7 +58,8 @@ def get_matches(self, video):
         if isinstance(video, Episode) and self.movie_kind == 'episode':
             # tag match, assume series, year, season and episode matches
             if self.matched_by == 'tag':
-                matches |= {'series', 'year', 'season', 'episode'}
+                if not video.imdb_id or (video.imdb_id and self.movie_imdb_id != video.imdb_id):
```
Diaoul: Shouldn't that be `==`?

Diaoul:
Please add a unittest.

subliminal/providers/opensubtitles.py
fernandog: @Diaoul i think `video.imdb_id` should be `video.series_imdb_id` right ?

tests/test_opensubtitles.py
```diff
@@ -260,3 +260,15 @@ def test_download_subtitle(movies):
...
+@pytest.mark.integration
+@vcr.use_cassette
+def test_tag_match(episodes):
+    video = episodes['the fall']
+    languages = {Language('por', 'BR')}
+    unwanted_subtitles = {'1954369181'}
+    with OpenSubtitlesProvider() as provider:
+        subtitles = provider.list_subtitles(video, languages)
+    assert len(subtitles) > 0
+    assert {subtitle.id for subtitle in subtitles} != unwanted_subtitles
```
fernandog: @Diaoul not sure if is the best unittest. Its the first time i did one. so please bear with me.

My goal is to test that: this wrong subtitle `1954369181` shouldn't be in the list of results because even if is a tagmatch, the imdb is wrong.

fernandog:
@Diaoul rebased, added missing cassete as requested

fernandog:
@Diaoul merged develop here. All requested changed were added

Diaoul:
LGTM

subliminal/providers/opensubtitles.py
```diff
@@ -60,7 +60,8 @@ def get_matches(self, video):
         if isinstance(video, Episode) and self.movie_kind == 'episode':
             # tag match, assume series, year, season and episode matches
             if self.matched_by == 'tag':
-                matches |= {'series', 'year', 'season', 'episode'}
+                if not video.imdb_id or (video.imdb_id and self.movie_imdb_id == video.imdb_id):
```
Diaoul: Could be simplified as

```python
if not video.imdb_id or video.imdb_id == self.movie_imdb_id:
```

fernandog: Ok! Thanks

PR #713, opened 2016-11-07
Dont stop on parse error.

Hellowlol:


Diaoul:
Travis is failing.

Hellowlol:
I have no idea why it fails it a really small change and I fail to see why this pr should make it fail

Diaoul:
Replace both `os.path` and `os.walk` imports by a simpler `import os`

Hellowlol:
What can i do to get this merged?

Diaoul:
ValueError is quite broad, I think it's best to restrict the scope of the exception handling to the only portion of the code that is problematic.

Hellowlol:
Valueerror was the exception i got. Ill try to run it again and see if can make it less broad

fernandog:
@Hellowlol any news on this one? I'm helping Diaoul to maintain v2 so as soon you fix the review we can merge

fernandog:
@Diaoul This is the exception from datetime:
ValueError: year is out of range

I don't think we can make less broad. What do you think?

Diaoul:
Then do something like this:
```python
# skip old files
try:
    file_age = datetime.utcfromtimestamp(os.path.getmtime(filepath))
except ValueError:
    logger.warning('Could not get age of file %r in %r', filename, dirpath)
else:
    if age and datetime.utcnow() - file_age > age:
        logger.debug('Skipping old file %r in %r', filename, dirpath)
        continue
```

fernandog:
Closing this as it was cherry-picked to another PR

PR #743, opened 2017-03-31
Fix empty archive's name in legendastv provider

hpsbranco:
When logged in as a donor, there's an extra `<a href=... class="seta"></a>` in the div, causing `get_archives` to return an archive with no name. This will result in lots of unnecessary (and useless) downloads when `download_best_subtitles` is called.

To fix we just specify the correct anchor (the one inside the `<p>`) in `get_archives`. Easy fix, ugly test (without sharing my password, that is).


hpsbranco:
stevedore-1.2.0 broke the RegistrableExtensionsManager.

It's an easy fix, I'll submit later.

hpsbranco:
The tests for non-donors were already in place.

I've seen you closed this and merged. Thank you.

fernandog:
@hpsbranco that wasn't a fix in subliminal, it was my alternative fix. Did you tested mine?

PR #714, opened 2016-11-10
Hotfix/subscenter url fix

ofir123:

PR #760, opened 2017-05-27
Fix subscenter provider

sagilo:
Tested and working


sagilo:
Almost all providers fail the tests, it's clear the master itself doesn't pass, is this project discontinued ?

Diaoul:
It is not although I admit I have not had much time lately to work on it. A few changes may be required here and there but no big deal.

Diaoul:
PEP8 doesn't pass here. Also, what does `h_version` mean?

subliminal/providers/subscenter.py
```diff
@@ -201,7 +203,7 @@ def query(self, title, season=None, episode=None):
 
                         # otherwise create it
                         subtitle = SubsCenterSubtitle(language, hearing_impaired, page_link, title, season, episode,
-                                                      title, subtitle_id, subtitle_key, downloaded, [release])
+                                                      title, subtitle_id, subtitle_key, downloaded, [release], h_version)
```
Diaoul: Line too long

subliminal/providers/subscenter.py
```diff
@@ -191,6 +192,7 @@ def query(self, title, season=None, episode=None):
                         subtitle_key = subtitle_item['key']
                         downloaded = subtitle_item['downloaded']
                         release = subtitle_item['subtitle_version']
+                        h_version = subtitle_item['h_version']
```
Diaoul: What is `h_version`?

subliminal/providers/subscenter.py
sagilo: `h_version` is an identifier for the subtitle provided by the provider, see example : 
http://www.subscenter.org/he/cst/data/series/sb/greys-anatomy/13/22/

subliminal/providers/subscenter.py
sagilo: It's some kind of a subtitle unique identifier which required as a param to get the subtitle.
See example here :
http://www.subscenter.org/he/cst/data/series/sb/greys-anatomy/13/22/

sagilo: @Diaoul I've updated the commit and the PEP8 test passes.
The `test_subscenter` looks obsolete and needs a lot of adjustments to pass (it doesn't look like it would pass with the current version as well but I'm speculating here)

Diaoul: Well, if you can make it so the test is accurate with regards to the current version that'd be great.

sagilo: @Diaoul It looks fine, the `test_download_subtitle` fails in python3.x.
It looks like the test depends on ordering within the Subtitle object and somewhere it is not the same between Python 2.x to Python 3.x (I've used python 3.5 to test and it passed with the same value as in Python 2.x)
Any ideas ?

sagilo: Fixed that too, latest test build seems to run with no internet connectivity.

sagilo: @Diaoul It's very unfortunate you ask people to do stuff and then ignores it.

Diaoul: I'm not ignoring it, just had little time to review this lately and have no notification that you pushed new commits to this PR. Meanwhile I noticed this other PR too: #714. Does that help? Or did you manage to get the code and tests working?

sagilo:
@Diaoul All subscenter tests pass, 4 other don't but they don't seem related.

PR #764, opened 2017-06-19
When ServiceUnavailable, log error instead of exception (traceback)

fernandog:
@Diaoul imho 503 status is not an unexpected error (exception). It's quite common btw. This PR avoids showing a traceback to the user and its show an error log message instead  `Provider {provider} unavailable, discarding it`

Also OpenSubtitle raises a xmlrpclib.ProtocolError when unavailable

Full traceback is in the commit message.
What do you think?


subliminal/core.py
```diff
@@ -703,3 +713,12 @@ def save_subtitles(video, subtitles, single=False, directory=None, encoding=None
...
     return saved_subtitles
+
+
+def raise_for_status(r):
+    if r.status_code == 304:
+        raise TooManyRequests()
+    elif r.status_code == 503:
+        raise ServiceUnavailable
+    else:
+        r.raise_for_status()
```
Diaoul: I agree this is not ideal to have unavailability errors go straight up as unhandled exceptions.

`r.status_code == 503` already raises HTTPError with response attribute on which you can check the status code
```python
try:
    r.raise_for_status()
except Exception as e:
    # e.reponse.status_code == 503
```
`TooManyRequests` is quite provider specific, I'm not sure this is relevant to put this in a global function.
Sometimes `raise_for_status` isn't used because one expects a redirect or something like that.

I suggest a less invasive approach: catch HTTPError with 503 status code pretty much as ServiceUnavailable and leave the TooManyRequests handling on a provider basis.
What do you think?

fernandog: @Diaoul thanks for reviewing. 
I added TooManyRequests only because of Addi7ed. I will add a custom to the provider then

Just a fyi, I did all this based on OpenSubtitles:
https://github.com/Diaoul/subliminal/blob/develop/subliminal/providers/opensubtitles.py#L268-L294

should we change that method too as it checks 503 and others status codes?

Diaoul: OpenSubtitles does that because there is no `raise_for_status` available. I agree that ServiceUnavailable shouldn't be related to OpenSubtitles though but maybe a broader exception class within whole subliminal.

subliminal/core.py
```diff
@@ -83,8 +83,11 @@ def __delitem__(self, name):
             logger.error('Provider %r timed out, improperly terminated', name)
         except (ServiceUnavailable, ProtocolError):  # OpenSubtitles raises xmlrpclib.ProtocolError when unavailable
             logger.error('Provider %r unavailable, improperly terminated', name)
-        except:
-            logger.exception('Provider %r terminated unexpectedly', name)
+        except Exception as e:
+            if hasattr(e, 'response') and getattr(e.response, 'status_code', None) == 503:
+                logger.error('Provider %r terminated unexpectedly', name)
+            else:
+                logger.exception('Provider %r terminated unexpectedly', name)
```
fernandog: @Diaoul like this?

Diaoul: Better to catch HTTPError from requests no?

subliminal/core.py
```diff
@@ -83,11 +83,13 @@ def __delitem__(self, name):
...
-            if hasattr(e, 'response') and getattr(e.response, 'status_code', None) == 503:
-                logger.error('Provider %r terminated unexpectedly', name)
+        except requests.HTTPError as e:
+            if e.response and e.response.status_code == 503:
+                logger.error('Provider %r unavailable, improperly terminated', name)
             else:
-                logger.exception('Provider %r terminated unexpectedly', name)
+                logger.exception('Provider %r http error, terminated unexpectedly', name)
+        except:
+            logger.exception('Provider %r terminated unexpectedly', name)
```
fernandog: @Diaoul all good?

subliminal/core.py
```diff
@@ -79,6 +81,13 @@ def __delitem__(self, name):
             self.initialized_providers[name].terminate()
         except (requests.Timeout, socket.timeout):
             logger.error('Provider %r timed out, improperly terminated', name)
+        except (ServiceUnavailable, ProtocolError):  # OpenSubtitles raises xmlrpclib.ProtocolError when unavailable
+            logger.error('Provider %r unavailable, improperly terminated', name)
+        except requests.HTTPError as e:
+            if e.response and e.response.status_code == 503:
+                logger.error('Provider %r unavailable, improperly terminated', name)
+            else:
+                logger.exception('Provider %r http error, terminated unexpectedly', name)
```
fernandog: @Diaoul all good?

fernandog:
@Diaoul can we make TooManyRequests from Addi7ed a "expected error" and also log as unavailable instead of log exception? It raises that error a lot in my tests with this PR

```
ERROR:subliminal.core:Unexpected error in provider 'addic7ed'
  File "build/bdist.linux-x86_64/egg/subliminal/core.py", line 118, in list_subtitles_provider
    return self[provider].list_subtitles(video, provider_languages)
  File "build/bdist.linux-x86_64/egg/subliminal/providers/addic7ed.py", line 273, in list_subtitles
    return [s for s in self.query(video.series, video.season, video.year)
  File "build/bdist.linux-x86_64/egg/subliminal/providers/addic7ed.py", line 238, in query
    raise TooManyRequests()
```

pannal:
Sure. I think we definitely need a more fine grained provider status handling than what's currently implemented. Is 304 on addic7ed really their "too many requests" response?

fernandog:
@pannal this are mine findings using a modified pytest:

```
self = <Addic7edProvider [(<class 'subliminal.video.Episode'>,)]>
series = 'The Big Bang Theory', season = 7, year = 2007, country = None

    def query(self, series, season, year=None, country=None):
        # get the show id
        show_id = self.get_show_id(series, year, country)
        if show_id is None:
            logger.error('No show id found for %r (%r)', series, {'year': year, 'country': country})
            return []
    
        # get the page of the season of the show
        logger.info('Getting the page of show id %d, season %d', show_id, season)
        r = self.session.get(self.server_url + 'show/%d' % show_id, params={'season': season}, timeout=10)
        print r
        raise_for_status(r)
        soup = ParserBeautifulSoup(r.content, ['lxml', 'html.parser'])
    
        # loop over subtitle rows
>       match = series_year_re.match(soup.select('#header font')[0].text.strip()[:-10])
E       IndexError: list index out of range

..\subliminal\providers\addic7ed.py:239: IndexError
---------------------------- Captured stdout call -----------------------------
<Response [304]>
---------------------------- Captured stderr call -----------------------------
No handlers could be found for logger "bs4.dammit"
========================== 1 failed in 59.79 seconds ==========================
Process finished with exit code 0
```

fernandog:
Travis failing only for Addi7ed using lxml parser (known issue)

Diaoul:
I don't like having the `raise_for_status` as a function as it only works for providers like requests and most importantly it makes adding new providers more difficult as you cannot just `r.raise..` but have to use the `utils.raise_for_status`. I think the provider code should be as simple as possible, we can catch the requests Exception and check the status_code from within the exception handler instead. This would keep the initial intentions of this PR and reduce the footprint. What do you think?

fernandog:
https://github.com/pymedusa/Medusa/issues/3121

PR #765, opened 2017-06-19
Fix RegistrableExtensionManager not working with stevedore after 1.19.1

fernandog:
@Diaoul stevedore 1.20.0 broke the test_extensions.py 

http://git.openstack.org/cgit/openstack/stevedore/commit/?id=eb04fc4a3e3db913475d2f7105420351a0a781a9

With 1.19.1 it passes

UPDATE: new version has a new method so my fix is to override both methods

I changed the first test because in newest version, `RegistrableExtensionManager` will only have the extensions that was passed


fernandog:
Close after ratoaq2 PR

PR #767, opened 2017-06-21
LegendasTV: Detect if new releases were added to the subtitle pack (rar file)

fernandog:
Fixes the common situation where a new subtitle releases are added to the pack (rar file)

Did this together with @ratoaq2

### DIFF
--- a/subliminal/providers/legendastv.py
+++ b/subliminal/providers/legendastv.py
@@ -45,7 +45,7 @@ rating_re = re.compile(r'nota (?P<rating>\d+)')
 timestamp_re = re.compile(r'(?P<day>\d+)/(?P<month>\d+)/(?P<year>\d+) - (?P<hour>\d+):(?P<minute>\d+)')
 
 #: Cache key for releases
-releases_key = __name__ + ':releases|{archive_id}'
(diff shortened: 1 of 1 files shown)


ratoaq2:
Legendas.tv website often updates the rar files with more subtitles for more releases for a given title.

Subliminal uses the "last update" timestamp to know if the cache should be discarded or not. But we found out that for several situations, legendas.tv website updates the rar file without updating the "last update" timestamp (argh!). But they update the archive name to include the new words for the added releases.

So, if the name is part of the cache key, subliminal won't get a cache hit when that happens, and the new subtitle will be detected and evaluated.

subliminal/providers/legendastv.py
```diff
@@ -387,7 +387,8 @@ def query(self, language, title, season=None, episode=None, year=None):
                 expiration_time = (datetime.utcnow().replace(tzinfo=pytz.utc) - a.timestamp).total_seconds()
 
                 # attempt to get the releases from the cache
-                releases = region.get(releases_key.format(archive_id=a.id), expiration_time=expiration_time)
+                cache_item = releases_key.format(archive_id=a.id, archive_name=a.name)
```
ratoaq2: probably you should name it: `cache_key` or something similar

Issue #707, opened 2016-10-22
Every season premiere could take up to 3 weeks to get subs using legendastv provider


ratoaq2:
For the latest subliminal version (2.0.5) or latest development version

Given any Series that has a season premiere (e.g: Season 3, episode 1) today/tonight, subliminal can take up to 3 weeks to download subtitles for the new episodes in the new season.

This part:
https://github.com/Diaoul/subliminal/blob/master/subliminal/providers/legendastv.py#L202-L203

searches for titles with a given name. In case of tv shows, each item of the returned list represents a "Season" for the given series. E.g.:
- My Show Season 1
- My Show Season 2

Since the returned list is cached, I noticed a delay of weeks for subliminal to identify a new "season entry" in that list and start searching for subtitles for the new episodes.

That method searches legendastv website using the very same endpoint that the auto suggestion input field "Procurar" ("Search" in English) in the right corner of the home page uses ( http://legendas.tv/ ). Every user that visits the home page and start typing, is actually hitting that endpoint. Its performance and response time is considerable good.

In my point of view, either the cache expiration time should be reduced or it should be dropped for that method.

Diaoul:
Show search should be cached. You cannot drop the cache for events that occur once in a while. What I suggest however is that applications using subliminal shoud force cache renewal whenever they feel this is necessary. For example, adding a newly released show is a good enough reason for forcing cache renewal.

Performance is not the purpose of the cache, it's a side effect. The purpose is to avoid hammering websites whenever doing a subtitle search. Web scrapping suffers bad reputation already, without the cache you would be hammering the website with tons of unnecessary requests.

I agree on the fact that 3 weeks is obviously overkill. This should be reduced to a more reasonable value.

PR #774, opened 2017-06-27
LegendasTV: Fix not finding news seasons for shows (each season is a title)

fernandog:
For episodes each season is a new title, so we can't cache only the title because that title it associated to a specific season, so I pass season in the `search_titles`. 
Also this method will only return valid titles, so we will only cache titles associated to that given season

Again with support from @ratoaq2 

UPDATE: 
I had to update the cassettes because the sanitize is now inside the `search_titles`
Two tests doesn't make sense now as search_titles will only return valid titles: both are mismatch on title
Updated tests to check only the valid titles


ratoaq2:
This should fix https://github.com/Diaoul/subliminal/issues/707

fernandog:
Rebased with latest develop. Ready to be reviewed

Issue #688, opened 2016-09-10
Provider is discarded because of one corrupted file


ratoaq2:
```
commit 7eb7a53fe6bcaf3e01a6b44c8366faf7c96f7f1b
Author: Antoine Bertin <diaoulael@gmail.com>
Date:   Sat Sep 3 21:54:47 2016 +0200

    Back to development version
```

LegendasTV Provider is discarded because of one corrupted file:

```
> subliminal --debug --legendastv user pass download -f -l pt-BR -p legendastv -m 97 Once.Upon.a.Time.S01E12.720p.HDTV.X264-DIMENSION.mkv
DEBUG:subliminal.cli:Collecting path Once.Upon.a.Time.S01E12.720p.HDTV.X264-DIMENSION.mkv

INFO:subliminal.providers.legendastv:Downloading archive 03ae4d4d7b3fb193ce58ca4abcb946dc
DEBUG:subliminal.providers.legendastv:Identified rar archive
ERROR:subliminal.core:Unexpected error in provider 'legendastv', discarding it
Traceback (most recent call last):
  File "/home/user/workspace/git/subliminal/subliminal/core.py", line 171, in download_subtitle
    self[subtitle.provider_name].download_subtitle(subtitle)
  File "/home/user/workspace/git/subliminal/subliminal/providers/legendastv.py", line 448, in download_subtitle
    subtitle.content = fix_line_ending(subtitle.archive.content.read(subtitle.name))
  File "/home/user/venv/subliminal/local/lib/python2.7/site-packages/rarfile.py", line 629, in read
    return f.read()
  File "/home/user/venv/subliminal/local/lib/python2.7/site-packages/rarfile.py", line 1369, in read
    self._check()
  File "/home/user/venv/subliminal/local/lib/python2.7/site-packages/rarfile.py", line 1384, in _check
    raise BadRarFile("Corrupt file - CRC check failed: " + self.inf.filename)
BadRarFile: Corrupt file - CRC check failed: Once.Upon.a.Time.S01E12.720p.HDTV.X264-DIMENSION.srt
WARNING:subliminal.core:Provider 'legendastv' is discarded
WARNING:subliminal.core:Provider 'legendastv' is discarded
WARNING:subliminal.core:Provider 'legendastv' is discarded
WARNING:subliminal.core:Provider 'legendastv' is discarded
WARNING:subliminal.core:Provider 'legendastv' is discarded
WARNING:subliminal.core:Provider 'legendastv' is discarded
[...]

PR #781, opened 2017-07-01
Don't discard provider when BadRarFile or BadZipFile

fernandog:
Show a 'Invalid subtitle' warning instead
Without this, until subtitle is manually downloaded, it will keep discarding provider every run

Fixes https://github.com/Diaoul/subliminal/issues/688
ping @ratoaq2


ratoaq2:
You could get badzipfile in several providers, not only in legendastv. The solution needs to be a general solution

PR #727, opened 2017-01-06
Better archive scanning with rarfile, avoids a lockup, finds right video file in archive with samples/intros

miigotu:
Signed-off-by: miigotu <miigotu@gmail.com>

PR #783, opened 2017-07-04
Better archive scanning with rarfile v2

fernandog:
"Better archive scanning with rarfile, avoids a lockup, finds right video file in archive with samples/intros"
Cherry-picked first commit from https://github.com/Diaoul/subliminal/pull/727
ping @miigotu as original author

Plus:
- Feature branch
- Fixed pytest
- Created rar methods so scan_archive can be simpler
- Added more tests

btw I can't rebase without loosing the original commit in cherry-pick


subliminal/core.py
```diff
@@ -357,6 +357,49 @@ def search_external_subtitles(path, directory=None):
     return subtitles
 
 
+def is_valid_file(path):
```
ratoaq2: This is misleading. The name of the method leads anyone to think that it returns true or false and the method implemented doesn't behave like that

subliminal/core.py
```diff
@@ -357,6 +357,49 @@ def search_external_subtitles(path, directory=None):
...
+    file_info.sort(key=operator.attrgetter('file_size'), reverse=True)
+
+    # no video found
+    if not file_info:
+        raise ValueError('No video in archive')
+
+    # Free the information about irrelevant files before guessing
+    file_info = file_info[0]
+
+    return file_info.filename, file_info.file_size
```
ratoaq2: The same here. The method name leads you to think that a video object is returned, but instead of that a tuple is returned.

subliminal/core.py
```diff
@@ -506,7 +530,7 @@ def scan_videos(path, age=None, archives=True):
             elif archives and filename.endswith(ARCHIVE_EXTENSIONS):  # archive
                 try:
                     video = scan_archive(filepath)
-                except (NotRarFile, RarCannotExec, ValueError):  # pragma: no cover
+                except (NotRarFile, RarCannotExec, Error, ValueError):  # pragma: no cover
                     logger.exception('Error scanning archive')
```
ratoaq2: Is there a reason for also catch Error?

subliminal/video.py
```diff
@@ -13,7 +13,7 @@
                     '.bix', '.box', '.cam', '.dat', '.divx', '.dmf', '.dv', '.dvr-ms', '.evo', '.flc', '.fli',
                     '.flic', '.flv', '.flx', '.gvi', '.gvp', '.h264', '.m1v', '.m2p', '.m2ts', '.m2v', '.m4e',
                     '.m4v', '.mjp', '.mjpeg', '.mjpg', '.mkv', '.moov', '.mov', '.movhd', '.movie', '.movx', '.mp4',
-                    '.mpe', '.mpeg', '.mpg', '.mpv', '.mpv2', '.mxf', '.nsv', '.nut', '.ogg', '.ogm' '.ogv', '.omf',
+                    '.mpe', '.mpeg', '.mpg', '.mpv', '.mpv2', '.mxf', '.nsv', '.nut', '.ogg', '.ogm', '.ogv', '.omf',
```
ratoaq2: This is related to something else, not rar file scanning

tests/test_core.py
```diff
@@ -632,3 +627,25 @@ def test_save_subtitles_single_directory_encoding(movies, tmpdir):
     path = os.path.join(str(tmpdir), os.path.splitext(os.path.split(movies['man_of_steel'].name)[1])[0] + '.srt')
     assert os.path.exists(path)
     assert io.open(path, encoding='utf-8').read() == u'ハローワールド'
+
+
+def test_scan_archive_with_no_video():
+    rar = os.path.realpath(os.path.join('tests', 'data', 'rar', 'README.rar'))
```
ratoaq2: What's inside of the real rar files? Are they big? Is there a way to test without actually having them?

subliminal/video.py
fernandog: Its because miigotu commit. I can't rebase because commit is signed

tests/test_core.py
fernandog: I tried. Would need to patch a lof of stuff from rarfile and I couldn't monkeypatch some.
The infolist is the most complicated one and the is_rarfile is not a class function
IMO is not worth it.

subliminal/core.py
fernandog: this is from miigotu commit and I think this is the reason as his comment:
```
# raise an exception if the rar file is broken
# must be called to avoid a potential deadlock with some broken rars
```

subliminal/core.py
fernandog: what do you suggest for the name?

subliminal/core.py
fernandog: this is from miigotu commit and I think this is the reason as his comment:
```
# raise an exception if the rar file is broken
# must be called to avoid a potential deadlock with some broken rars
```

subliminal/core.py
fernandog: Fixed

subliminal/core.py
```diff
@@ -357,6 +357,48 @@ def search_external_subtitles(path, directory=None):
...
+
+    # no video found
+    if not file_info:
+        raise ValueError('No video in archive')
+
+    # Free the information about irrelevant files before guessing
+    file_info = file_info[0]
+    return file_info.filename, file_info.file_size
+
+
```
Diaoul: It seems those two functions are called only once so there is no added value in creating separate functions, just put that in the `scan_video`, it will save some trouble for code readers to go back and forth from the function definition and the main `scan_video`.

subliminal/core.py
```diff
@@ -357,6 +357,48 @@ def search_external_subtitles(path, directory=None):
...
 
+def raise_for_invalid_file(path):
+    """Checks if is a valid file."""
+    # check for non-existing path
+    if not os.path.exists(path):
+        raise ValueError('Path does not exist')
+
+    # check that this is a file and not a directory
+    if not os.path.isfile(path):
+        raise ValueError('%r is not a file' % path)
```
Diaoul: You shouldn't ever hit this exception unless you explicitly call `scan_video` on a directory. If you do that, don't expect the `scan_video` function to work properly. We are all consenting adults here

fernandog:
TODO:
Need to add tests using fixture as requested. I need sometime as I never used fixture. Need to figure it out

sharkykh:
> btw I can't rebase without loosing the original commit in cherry-pick

If I understand you correctly, you can rebase,
by using `rebase -i master`, mark that commit with `edit`, and then use `git commit --amend -c COMMIT_HASH`. This will use the info from that commit, like message, author and date.

fernandog:
@Diaoul i was unable to to as you requested: `fake rarfile with a fixture `. Need your help

ratoaq2:
PR replaced by #953

PR #694, opened 2016-09-28
Feature: Add Opensubtitle trusted option

fernandog:
@Diaoul the only thing i couldn't figure it out in `click.option` is how to accept only 2 params for opensbutitles and considers trusted as 'no' if not passed (to be compatible with existing users scripts)

command: `subliminal --debug --opensubtitles user password yes download -l pt-BR -p opensubtitles -m 100`

Partial log:
```
DEBUG:subliminal.providers.opensubtitles:Found subtitle <OpenSubtitlesSubtitle '1955321728' [pt-BR]> by moviehash
DEBUG:subliminal.providers.opensubtitles:Found subtitle <OpenSubtitlesSubtitle '1955321728' [pt-BR]> by tag
DEBUG:subliminal.providers.opensubtitles:Found subtitle <OpenSubtitlesSubtitle '1955321728' [pt-BR]> by fulltext
DEBUG:subliminal.providers.opensubtitles:Discarding found subtitle id 1955321122 as it is not uploaded by trusted user
DEBUG:subliminal.providers.opensubtitles:Found subtitle <OpenSubtitlesSubtitle '1955321720' [pt-BR]> by fulltext
```

Full log:
```
subliminal --debug --opensubtitles user password yes download -l pt-BR -p opensubtitles -m 100 "/series/Marvel's Agents of S.H.I.E.L.D/Season 4/Marvels.Agents.of.S.H.I.E.L.D.S04E01.The.Ghost.1080p.WEB-DL.DD5.1.H.264-AG.mkv"
Collecting videos  [####################################]  100%  /series/Marvel's Agents of S.H.I.E.L.D/Season 4/Marvels.Agents.of.S.H.I.E.L.D.S04E01.The.Ghost.1080p.WEB-DL.DD5.1.H.264-AG.mkvDEBUG:subliminal.cli:Collecting path /series/Marvel's Agents of S.H.I.E.L.D/Season 4/Marvels.Agents.of.S.H.I.E.L.D.S04E01.The.Ghost.1080p.WEB-DL.DD5.1.H.264-AG.mkv
INFO:subliminal.core:Scanning video 'Marvels.Agents.of.S.H.I.E.L.D.S04E01.The.Ghost.1080p.WEB-DL.DD5.1.H.264-AG.mkv' in "/series/Marvel's Agents of S.H.I.E.L.D/Season 4"
DEBUG:subliminal.core:Size is 1790193917
[...]

PR #785, opened 2017-07-05
Add trusted subtitle uploader option

fernandog:
@Diaoul this is a **initial idea** as discussed here: https://github.com/Diaoul/subliminal/pull/694#issuecomment-250394866

- Adds new param "-t" to only download from trusted subtitle uploaders
- Added for OpenSubtitles and LegendasTV
- Added tests

If trusted mode is "disabled" then current max score will still be 359
If trusted mode is "enabled" then max score will be 459

I choose 100 for trusted so it can be easier to "read" scores
Sure I can change it

At least to me i would like to only download from trusted uploaders

But I see that some users may want not-trusted uploaders so the trusted option could be like `trusted_only`,  `prefer_trusted` so:
trusted_only : only download from trusted
prefer_trusted: will download sub if matches score even if not trusted sub available


subliminal/providers/legendastv.py
```diff
@@ -469,7 +474,7 @@ def query(self, language, title, season=None, episode=None, year=None):
                 # iterate over releases
                 for r in releases:
                     subtitle = self.subtitle_class(language, t['type'], t['title'], t.get('year'), t.get('imdb_id'),
-                                                   t.get('season'), a, r)
+                                                   t.get('season'), a, r, a.featured)
```
Diaoul: If it is named featured here it should be named featured in the Subtitle class too.

fernandog:
@Diaoul how to proceed then?

Diaoul:
I don't have the answer to this question. I think we need to differentiate how a subtitle matches a video and what user would prefer. The compute_score function is becoming a sort of mix between the two which is not a good thing.

PR #787, opened 2017-07-07
Napisy24 provider

### LINKED ISSUE #742: Napisy24 provider
Any chance to add napisy24.pl as a provider?
There is no public API, but [qnapi](https://github.com/QNapi/qnapi) project has a [working solution](https://github.com/QNapi/qnapi/blob/master/libqnapi/src/engines/qnapisy24engine.cpp).

MoshiMoshi0:
Support for Napisy24 provider but without proper handling of subtitles in formats other than .srt. Waiting for Diaoul's converter.

The hash algorithm of Napisy24 is the same as opensubtitles but I've added separate [hash function](https://github.com/devcompl/subliminal/commit/153d59bbfb14be6e4288acd546bcdc0f208c7b03#diff-5a14a4344c1cdae78b36956ceb6b6e6c).

Closes #742.


wujekbogdan:
Why was this PR never merged? I could try to revive it if there's a chance it gets merged into core.

Issue #652, opened 2016-04-25
LTV: filename and subtitle are "file match" but no release group in filename and subtitle


fernandog:
As there is no 'release group' in both filename and subtitle it doesn't match release group and score get two low

Maybe in LTV when subtitle name == filename we make return a score 215 (hash) ?
or another solution?

Release: Castle.S08E18.Backstabber.1080p.WEB-DL.DD5.1.H.264.mkv (no release group)

```
Downloading subtitles  [####################################]  100%  Castle.S08E18.Backstabber.1080p.WEB-DL.DD5.1.H.264.mkv
INFO:subliminal.core:Listing subtitles with provider 'legendastv' and languages set([<Language [pt-BR]>])
DEBUG:subliminal.providers.legendastv:Found subtitle <LegendasTVSubtitle u'5719b8d41ad54-castle.s08e18.backstabber.1080p.web-dl.dd5.1.h.264.srt' [pt-BR]>
INFO:subliminal.score:Computing score of <LegendasTVSubtitle u'5719b8d41ad54-castle.s08e18.backstabber.1080p.web-dl.dd5.1.h.264.srt' [pt-BR]> for video <Episode [u'Castle', 2003, 8x18]> with {'hearing_impaired': False}
INFO:subliminal.score:Computed score 204 with final matches set(['episode', 'format', 'series', 'year', 'season', 'video_codec', 'resolution'])
```

@ratoaq2
##

Diaoul:
Too low for what? If the video doesn't carry the release group, it'll be the highest non-hash match for that video, no need to fake the hash match.
If you're not happy with the score computation you can implement your own, it's as simple as overriding a function for library users.

fernandog:
too low for "perfect matches". the min_score we use is : 

```
    if sickbeard.SUBTITLES_PERFECT_MATCH:
        return episode_scores['hash'] - (episode_scores['resolution'] +
                                         episode_scores['video_codec'] +
                                         episode_scores['audio_codec'])
```

its just a suggestion. maybe can be improved in other way.

Diaoul:
What you can do is scale down the hash to the maximum score possible (sum of scores from filename tokens).
However, in case of really poor filename, hash will be better but with this it won't.

ratoaq2:
I was taking a look at this question once more... and I realized one thing:
If you decide to build a custom score computation then you don't have direct access to guess dict that used in the actual logic:

[...]

PR #789, opened 2017-07-11
Match release_group when both video and subtitle doesn't have it

fernandog:
This is a proposal to fix my issue:
https://github.com/Diaoul/subliminal/issues/652

If both subtitles and video doesn't have release_group, we assume is a match on release_group

@ratoaq2 suggested this: https://github.com/Diaoul/subliminal/issues/652#issuecomment-244585101 that we can consider too

What do you think @Diaoul ?
If you don't like it at least we can discuss alternative solutions


ratoaq2:
I don't agree with my proposal from 10 months ago since I have a more mature opinion on that. 

From my point of view the compute score function should receive more information, not only a set of which fields have matched. 
Maybe it should be a dict with matched fields as key and the actual value as values. 
Then the score function can decide how to score matches with or without values

Diaoul:
I tend to agree with @ratoaq2, there shouldn't be any logic in `guess_matches`, it should only parse tokens so that `compute_score` can do its job.

fernandog:
@Diaoul Its a big change in code right?

Diaoul:
The video object and the subtitle object are already passed to the compute_score function.

Anyway, if the release_group isn't present in the video name it means it has probably been renamed, matching the absence of release_group isn't a good idea here. It may suit your use case but I can see side effects.

duramato:
>> ...if the release_group isn't present in the video name it means it has probably been renamed...

Incorrect, there are many _releases_ that don't have a release group in them because they where released like that.

PR #762, opened 2017-06-16
LegendasTV: Use new page to search for titles

fernandog:
@Diaoul after 1 year heavily using LTV provider I'm proposing this fixes. They were fully tested in Medusa app

I appreciate if you can give me our feedback
thanks

PR #792, opened 2017-07-22
Fix broken pytest after changing provider search URL

fernandog:
Test was fine in the original commit but after I rebased with latest develop and this https://github.com/Diaoul/subliminal/pull/762/files got merged, I didn't updated cassette so test started to fail.
Thanks @sharkykh for finding this.

Sorry about that @Diaoul


Diaoul:
Keep a test for download.

Issue #709, opened 2016-10-29
can't get legendastv subtitles


evertonbob:
What am i doing wrong?
`
bobs-iMac:~ bob$ subliminal --debug --legendastv user pass download -f -v -l pt-BR -p legendastv The.Walking.Dead.S07E01.HDTV.x264-KILLERS`

This occurred with every single episode

[debug.txt](https://github.com/Diaoul/subliminal/files/559722/debug.txt)

Thanks.

```
INFO:subliminal.providers.legendastv:Downloading archive 580da91eac76b
DEBUG:subliminal.providers.legendastv:Identified rar archive
ERROR:subliminal.core:Unexpected error in provider 'legendastv', discarding it
Traceback (most recent call last):
  File "/Library/Python/2.7/site-packages/subliminal/core.py", line 171, in download_subtitle
    self[subtitle.provider_name].download_subtitle(subtitle)
  File "/Library/Python/2.7/site-packages/subliminal/providers/legendastv.py", line 448, in download_subtitle
    subtitle.content = fix_line_ending(subtitle.archive.content.read(subtitle.name))
  File "/Library/Python/2.7/site-packages/rarfile.py", line 629, in read
    return f.read()
  File "/Library/Python/2.7/site-packages/rarfile.py", line 1364, in read
    raise BadRarFile("Failed the read enough data")
BadRarFile: Failed the read enough data
WARNING:subliminal.core:Provider 'legendastv' is discarded
WARNING:subliminal.core:Provider 'legendastv' is discarded
WARNING:subliminal.core:Provider 'legendastv' is discarded
WARNING:subliminal.core:Provider 'legendastv' is discarded
WARNING:subliminal.core:Provider 'legendastv' is discarded
WARNING:subliminal.core:Provider 'legendastv' is discarded
WARNING:subliminal.core:Provider 'legendastv' is discarded
WARNING:subliminal.core:Provider 'legendastv' is discarded
WARNING:subliminal.core:Provider 'legendastv' is discarded
WARNING:subliminal.core:Provider 'legendastv' is discarded
WARNING:subliminal.core:Provider 'legendastv' is discarded
WARNING:subliminal.core:Provider 'legendastv' is discarded
WARNING:subliminal.core:Provider 'legendastv' is discarded
WARNING:subliminal.core:Provider 'legendastv' is discarded
WARNING:subliminal.core:Provider 'legendastv' is discarded
Downloading subtitles  [####################################]  100%                                           
[...]

PR #793, opened 2017-07-24
LegendasTV: Check if UNRAR tool is available and discard provider if not

fernandog:
As discussed here:
https://github.com/Diaoul/subliminal/issues/709


subliminal/providers/legendastv.py
```diff
@@ -164,6 +164,13 @@ class LegendasTVProvider(Provider):
     subtitle_class = LegendasTVSubtitle
 
     def __init__(self, username=None, password=None):
+
+        # Provider needs UNRAR installed. If not available raise ConfigurationError
+        try:
+            rarfile._check_unrar_tool()
```
sharkykh: I don't think it actually raises `rarfile.RarCannotExec`.
See [markokr/rarfile@rarfile_3_0-`/rarfile.py#L2927-L2948`](https://github.com/markokr/rarfile/blob/rarfile_3_0/rarfile.py#L2927-L2948)

And a note, in the next version, it will return True or False.
See https://github.com/markokr/rarfile/pull/32

fernandog: @sharkykh from the docs both seem valid:

Problem reported by unrar/rar.
http://rarfile.readthedocs.io/en/latest/api.html#rarfile.RarExecError

Executable not found.
http://rarfile.readthedocs.io/en/latest/api.html#rarfile.RarCannotExec

sharkykh: I didn't say they weren't valid, I said I don't think it raises that exception,
because look at the code in the first link.
```python
try:
    # does UNRAR_TOOL work?
    custom_check([ORIG_UNRAR_TOOL], True)
except RarCannotExec:
    try:
        # does ALT_TOOL work?
        custom_check([ALT_TOOL] + list(ALT_CHECK_ARGS), True)
    except RarCannotExec:
        # no usable tool, only uncompressed archives work
        pass
```

It passes, meaning - doesn't raise it.
It will only raise RarExecError which is the base exception for RarCannotExec.
I'm not sure how Python handles that.

tests/test_legendastv.py
```diff
@@ -333,3 +334,14 @@ def test_under_maintenance(movies):
             pass
         else:
             pytest.fail()
+
+
+@pytest.mark.integration
+def test_unrar_not_available(monkeypatch):
+    monkeypatch.setattr(rarfile, 'UNRAR_TOOL', 'test')
```
Diaoul: If this is supposed to be a path to the unrar binary maybe you can replace `test` by something more meaningful?

fernandog: @Diaoul goal was to set an invalid path. I will change it

Diaoul:
LGTM

fernandog:
@Diaoul ok to merge?

Diaoul:
Yep

PR #794, opened 2017-07-24
Podnapisi: discard earlier duplicate subtitle

fernandog:
@Diaoul 
I manage to reproduce issue 738:

```
INFO:subliminal.score:Computing score of <PodnapisiSubtitle 'PO9C' [fr]> for video <Movie ['Caf\xc3\xa9 Society', 2016]> with {'hearing_impaired': False}
DEBUG:subliminal.score:Using scores {'hash': 119, 'hearing_impaired': 1, 'audio_codec': 3, 'title': 60, 'video_codec': 2, 'format': 7, 'year': 30, 'resolution': 2, 'release_group': 15}
/usr/local/lib/python2.7/dist-packages/subliminal/providers/podnapisi.py:69: UnicodeWarning: Unicode equal comparison failed to convert both arguments to Unicode - interpreting them as being unequal
  if video.title and sanitize(self.title) == sanitize(video.title):
DEBUG:subliminal.score:Found matches set(['resolution', 'year'])
```

I used the same subtitle: PodnapisiSubtitle 'PO9C' [fr]>
Test pass *BUT* that UnicodeWarning is raised

**In the screenshot you can see we are comparing STR with UNICODE:**

How to proceed?

image


Diaoul:
This is not a fix for the mentioned issue, it just happens to fix the issue because this is a duplicate but what if it's not?

fernandog:
I didn't fix the issue. I only manage to reproduce. The duplicate change is just an improve. I asked how to proceed  to fix the issue

Diaoul:
I don't get why we have unicode for `self.title`. Should be str here.

fernandog:
I think here is already unicode:

`title = subtitle_xml.find('title').text`

There is no unicode conversion in the provider code
Not an expert so I dont know which is the best fix.

fernandog:
For some titles is unicode, for other str
No idea how to fix that.

@ratoaq2 you did some great changes about this in medusa, what do you suggest?

image

fernandog:
BTW provider doesn't accept "café society" as keyword. It returns 1500 results, most wrong results
https://podnapisi.net/subtitles/search/old?sK=café+Society&sXML=1&sL=fr

Searching for "cafe society" it returns only 4 valid results:
https://podnapisi.net/subtitles/search/old?sK=cafe+Society&sXML=1&sL=fr

fernandog:
@Diaoul here is the proposal to fix the issues:

- UnicodeWarning: Unicode equal comparison failed to convert both arguments to Unicode
- Provider doesn't allow keyword search to have accents

subliminal/providers/podnapisi.py
```diff
@@ -45,11 +45,12 @@ def id(self):
 
     def get_matches(self, video):
         matches = set()
-
+        # Make sure we don't compare unicode with str
+        title = self.title.encode("utf-8") if isinstance(self.title, unicode) else self.title
```
fernandog: about your comment: `I don't get why we have unicode for self.title. Should be str here.`. This is the implementation of it.

fernandog: @Diaoul should this be done for all providers?

fernandog:
@MasterMind2k can you give some insigh on this? Can you read everything and comment please?

MasterMind2k:
We have two searches, one is the old (the XML one) and the other is the new one. The new one is handling unicode properly (unfortunately it does not have machine readable output, yet!):
https://www.podnapisi.net/sl/subtitles/search/?keywords=café%2BSociety

I will check the code and provide a fix.

As for XML format. Do note that we print out in utf-8 format. Observe the header of the output
```
<?xml version="1.0" encoding="utf-8" standalone="yes"?>
```
So all text returned from our XML (and later JSON) will have to be treated as `unicode`.

fernandog:
@MasterMind2k ok but the search query can't be Unicode right? As one of my comments above

MasterMind2k:
Why not? It depends on the client tho.

You could do a percent encoding of non-ASCII characters tho. But, our search is the problem, not your client.

I'll respond when I'll fix the issue.

MasterMind2k:
It has been fixed and is now pushed on production.

Searching via old search API with unicodes is now available.

fernandog:
Thanks! I will update the PR soonish

fernandog:
@medariox can you try this changes in Medusa for a few days?

medariox:
Tested with and without unicode, all good 👍

PR #795, opened 2017-07-25
Fix lxml parsing for Addic7ed

sharkykh:
Basically, after like, 2 or 3 hours of trying different things,
I've come to the conclusion that,
  1. Addic7ed HTML markup is probably broken.
  2. Parsing using lxml worked just fine up until version lxml 3.7.0 was released.
      See [the last Travis job log for `PARSER=lxml` that passed](https://travis-ci.org/Diaoul/subliminal/jobs/169318251#L620).
    Something changed in lxml that caused it to not be able to parse Addic7ed's (broken) html properly, only parsing the some of the site.

Since we can't control Addic7ed's HTML code, we need to find a way around it.
(Oh, the joys of website scraping...)

Using a regex pattern is probably the simplest and fastest way to go about this, and it seems to work.
What it does is strip down the HTML to only the `<td class="version">` tags and discarding everything else.

Note: There are two commits, this is intentional, in order to show my progress.
- The first commit has a slower method - checking if it failed, and if it did then re-parsing with the stripped content.
This makes parsing the page with lxml much slower, **to a point where html.parser is faster**.
- The second commit just simplifies everything, and strips down the HTML **either way**.
    If the stripping down fails, it falls back to using the current method of BeautifulSoup-ing the whole response content **using only html.parser**, because we already know lxml can't parse it.
    This method is faster than both the first commit, and the current method, because BS4 only has to parse about half the HTML (`1.07 MB => 621 KB`).

Ping @fernandog

Some code I used for testing:
https://gist.github.com/sharkykh/40b059cfcdcf5aab360a3bcd7db65ed5

### DIFF
--- a/subliminal/providers/addic7ed.py
+++ b/subliminal/providers/addic7ed.py
@@ -19,6 +19,8 @@ logger = logging.getLogger(__name__)
 
 language_converters.register('addic7ed = subliminal.converters.addic7ed:Addic7edConverter')
 
+# Series cell matching regex
+show_cells_re = re.compile(r'<td class="version">.*?</td>', re.DOTALL)
 #: Series header parsing regex
 series_year_re = re.compile(r'^(?P<series>[ \w\'.:(),&!?-]+?)(?: \((?P<year>\d{4})\))?$')
 
@@ -138,9 +140,29 @@ class Addic7edProvider(Provider):
         r.raise_for_status()
         soup = ParserBeautifulSoup(r.content, ['lxml', 'html.parser'])
 
+        show_links = soup.select('td.version > h3 > a[href^="/show/"]')
+        if not show_links and soup.builder.NAME == 'lxml':
+            # LXML parser seems to fail when parsing Addic7ed.com HTML markup.
+            # Last known version to work properly is 3.6.4 (next version, 3.7.0, fails)
+            # This solution assumes that the site's markup is bad, and strips it down to only contain what's needed.
+            # It could fall back to html.parser whenever that fails,
+            # but that would mean not knowing it failed, and that's bad.
+            show_cells = re.findall(show_cells_re, r.content)
+            if show_cells:
+                # Re-parse stripped down HTML
+                soup = ParserBeautifulSoup(''.join(show_cells), ['lxml', 'html.parser'])
+            else:
+                # RegExp match failed
+                pass
+                # Re-parse using html.parser
+                # soup = ParserBeautifulSoup(r.content, ['html.parser'])
+
+            # Re-select elements and continue
+            show_links = soup.select('td.version > h3 > a[href^="/show/"]')
+
         # populate the show ids
         show_ids = {}
-        for show in soup.select('td.version > h3 > a[href^="/show/"]'):
+        for show in show_links:
             show_ids[sanitize(show.text)] = int(show['href'][6:])
         logger.debug('Found %d show ids', len(show_ids))
 
--- a/subliminal/providers/addic7ed.py
+++ b/subliminal/providers/addic7ed.py
@@ -2,6 +2,7 @@
 import logging
 import re
 
+import six
 from babelfish import Language, language_converters
 from guessit import guessit
 from requests import Session
@@ -20,7 +21,10 @@ logger = logging.getLogger(__name__)
 language_converters.register('addic7ed = subliminal.converters.addic7ed:Addic7edConverter')
 
 # Series cell matching regex
-show_cells_re = re.compile(r'<td class="version">.*?</td>', re.DOTALL)
+if six.PY2:
+    show_cells_re = re.compile(r'<td class="version">.*?</td>', re.DOTALL)
+else:
+    show_cells_re = re.compile(br'<td class="version">.*?</td>', re.DOTALL)
 #: Series header parsing regex
 series_year_re = re.compile(r'^(?P<series>[ \w\'.:(),&!?-]+?)(?: \((?P<year>\d{4})\))?$')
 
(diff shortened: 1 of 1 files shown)


sharkykh:
Well, my solution seems to raise a TypeError for Python 3.
```
______________________________ test_get_show_ids _______________________________
    @pytest.mark.integration
    @vcr.use_cassette('test_get_show_ids')
    def test_get_show_ids():
        with Addic7edProvider() as provider:
>           show_ids = provider._get_show_ids()
tests/test_addic7ed.py:182: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../virtualenv/python3.3.5/lib/python3.3/site-packages/dogpile/cache/region.py:1231: in decorate
    should_cache_fn)
../../../virtualenv/python3.3.5/lib/python3.3/site-packages/dogpile/cache/region.py:833: in get_or_create
    async_creator) as value:
../../../virtualenv/python3.3.5/lib/python3.3/site-packages/dogpile/lock.py:154: in __enter__
    return self._enter()
../../../virtualenv/python3.3.5/lib/python3.3/site-packages/dogpile/lock.py:94: in _enter
    generated = self._enter_create(createdtime)
../../../virtualenv/python3.3.5/lib/python3.3/site-packages/dogpile/lock.py:145: in _enter_create
    created = self.creator()
../../../virtualenv/python3.3.5/lib/python3.3/site-packages/dogpile/cache/region.py:800: in gen_value
    created_value = creator()
../../../virtualenv/python3.3.5/lib/python3.3/site-packages/dogpile/cache/region.py:1227: in creator
    return fn(*arg, **kw)
subliminal/providers/addic7ed.py:145: in _get_show_ids
    show_cells = re.findall(show_cells_re, r.content)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
pattern = <_sre.SRE_Pattern object at 0x2ac9ce0>
string = b'<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional....cript>\n                                                                                        \n\n</body>\n</html>\n'
flags = 0
    def findall(pattern, string, flags=0):
        """Return a list of all non-overlapping matches in the string.
    
        If one or more capturing groups are present in the pattern, return
        a list of groups; this will be a list of tuples if the pattern
        has more than one group.
    
        Empty matches are included in the result."""
>       return _compile(pattern, flags).findall(string)
E       TypeError: can't use a string pattern on a bytes-like object
```

sharkykh:
It seems [only 3.6 fails right now](https://travis-ci.org/Diaoul/subliminal/builds/257855700), but [it has something to do with SubsCenter](https://travis-ci.org/Diaoul/subliminal/jobs/257855709#L1038).
Possibly related to #714 (ping @ofir123).

fernandog:
@sharkykh can you please rebase as subscenter got removed?

sharkykh:
@fernandog Done

Diaoul:
If this is specific to lxml 3.7 can we enforce lxml < 3.7 as requirement for now ? Maybe this is a bug in lxml itself that will be fixed eventually.

sharkykh:
@Diaoul 
https://travis-ci.org/sharkykh/subliminal/jobs/259802067#L599
Now it fails with `lxml < 3.7` as well (no other fixes included, just changed lxml installed version)

The reason I said 3.6.4 was the last version to work was because of https://travis-ci.org/Diaoul/subliminal/jobs/169318251#L620.
I couldn't actually install 3.6.4 locally to test.
But I could install newer versions, since 3.7.0 (including), so I did.

sharkykh:
Just to be clear, the suggested solution for lxml still works after the rebase.
It fails all of the tests due to another issue (again).

fernandog:
@sharkykh can you rebase both branches please? The yaml was reverted.

sharkykh:
@fernandog Rebased this PR.
And [fix-lxml-v2 build #2 started](https://travis-ci.org/sharkykh/subliminal/builds/259954791)

fernandog:
@sharkykh Im my tests only lxml 3.6.0 passed. Anything higher than that failed. Can you set 3.6.0 in your branch?

sharkykh:
@fernandog 
https://travis-ci.org/sharkykh/subliminal/builds/259966309

sharkykh:
Nope. Still fails. This just got more complicated...

fernandog:
Agree. I think we should use this PR instead of enforce an old lxml version. What do you think @Diaoul ?

subliminal/providers/addic7ed.py
```diff
@@ -136,7 +139,16 @@ def _get_show_ids(self):
...
         r.raise_for_status()
-        soup = ParserBeautifulSoup(r.content, ['lxml', 'html.parser'])
+
+        # LXML parser seems to fail when parsing Addic7ed.com HTML markup.
+        # Last known version to work properly is 3.6.4 (next version, 3.7.0, fails)
+        # Assuming the site's markup is bad, and stripping it down to only contain what's needed.
+        show_cells = re.findall(show_cells_re, r.content)
+        if show_cells:
+            soup = ParserBeautifulSoup(b''.join(show_cells), ['lxml', 'html.parser'])
+        else:
```
sharkykh: Should I add
```python
# pragma: no cover
```
on this line?

PR #809, opened 2017-08-31
Add Hungarian provider Hosszupuska

morpheus133:


morpheus133:
Hi!
I see that test are runnig for lxml and native parser also. 
But it seems for me that in case of this provider the results are different for them. 
Which one I need to use?

subliminal/providers/hosszupuska.py
```diff
@@ -0,0 +1,233 @@
...
+    server_url = 'http://hosszupuskasub.com/'
+
+    def initialize(self):
+        self.session = Session()
+        # self.session.headers['User-Agent'] = 'Subliminal/%s' % __short_version__
+
+    def terminate(self):
+        self.session.close()
+
+    def GetLanguage(self, text):
```
ratoaq2: I see no reason to mix conventions and use CamelCase in two methods while all the others follow the expected snake_case convention

morpheus133: Thanks, for the comment. 
I corrected them.

PR #815, opened 2017-10-01
Update subscenter domain (.info)

sagilo:


fernandog:
PRed against master. should be develop. Also there is already works for subscenter

PR #817, opened 2017-10-02
Feature/new subscenter support

ofir123:
Fixed SubsCenter provider to work with the new API.


fernandog:
@ofir123 also http://www.subscenter.info/he/ still exists

ofir123:
This provider is based on the SubsCenter Kodi plugin.
I think they're the same (Cinemast is just an alternative name).. but of course I'll rename it if it's a problem.
subscenter still exists, but they added a CAPTCHA so automatic downloads are problematic.

fernandog:
@ofir123 ok but if site is Cinemast I would rename to not confuse users

ofir123:
Renamed to Cinemast!

fernandog:
@ofir123 can you merge develop branch into this branch? Or rebase on develop

ofir123:
Sure, rebased now.

fernandog:
When you rebase only your commits should appear

fernandog:
@ofir123 do this:

git checkout origin/develop
git pull
git checkout feature/new-subscenter-support
git rebase develop feature/new-subscenter-support
git push feature/new-subscenter-support -f

ofir123:
Oh sorry about that :\
Wrong direction.. fixed it now (and thanks for the tips @fernandog )

fernandog:
@ofir123 still has duplicate commits. I will try to fix tomorrow

fernandog:
@ofir123 fixed

fernandog:
@ofir123 can you add a test like legendastv that would only match a title if using an alternative title? Do you know one case that only return results when searching with alternative title?

ofir123:
@fernandog thnx for the fixes!
I've added the test you requested..

fernandog:
@ofir123 all tests passing

@Diaoul this is the the subscenter provider (which is in master) that got renamed to Cinemast. 
Good to merge?

subliminal/providers/cinemast.py
```diff
@@ -0,0 +1,231 @@
...
+    default_username = 'subliminal@gmail.com'
+    default_password = 'subliminal'
+
+    def __init__(self, username=None, password=None):
+        if any((username, password)) and not all((username, password)):
+            raise ConfigurationError('Username and password must be specified')
+
+        self.session = None
+        self.username = username or self.default_username
+        self.password = password or self.default_password
```
Diaoul: Why is there a default user?

subliminal/providers/cinemast.py
```diff
@@ -0,0 +1,231 @@
...
+        self.user_id = None
+        self.token = None
+        self.session = None
+
+    def initialize(self):
+        self.session = Session()
+        self.session.headers['User-Agent'] = 'Subliminal/{}'.format(__short_version__)
+
+        # login
+        if self.username and self.password:
```
Diaoul: This will always evaluate to `True` with default values.

subliminal/providers/cinemast.py
```diff
@@ -0,0 +1,231 @@
...
+        try:
+            results = r.json()
+        except ValueError:
+            return {}
+
+        # loop over results
+        subtitles = {}
+        for group_data in results.get('data', []):
+            # create page link
+            slug_name = group_data.get('name_en').lower().replace(' ', '-').replace('\'', '').replace('"', '')
```
Diaoul: What about other characters? e.g. dot, semicolons
Is there any other way to get this information?

subliminal/providers/cinemast.py
```diff
@@ -0,0 +1,231 @@
...
+            if query['type'] == 'series':
+                page_link = self.server_url + 'subtitle/series/{}/{}/{}/'.format(slug_name, season, episode)
```
Diaoul: urlencode is required here due to above comment.

subliminal/providers/cinemast.py
```diff
@@ -0,0 +1,231 @@
...
+            else:
+                page_link = self.server_url + 'subtitle/movie/{}/'.format(slug_name)
```
Diaoul: Same here

subliminal/providers/cinemast.py
```diff
@@ -0,0 +1,231 @@
...
+                    language = Language.fromalpha2(language_code)
+                    subtitle_id = subtitle_item['id']
+                    subtitle_key = subtitle_item['key']
+                    release = subtitle_item['version']
+
+                    # add the release and increment downloaded count if we already have the subtitle
+                    if subtitle_id in subtitles:
+                        logger.debug('Found additional release %r for subtitle %r', release, subtitle_id)
+                        bisect.insort_left(subtitles[subtitle_id].releases, release)  # deterministic order
+                        subtitles[subtitle_id].downloaded += 1
```
Diaoul: Why do you increase the downloaded count? Isn't that the same as `len(subtitles[subtitle_id].releases)`?

subliminal/providers/cinemast.py
```diff
@@ -0,0 +1,231 @@
...
+        season = episode = None
+
+        if isinstance(video, Episode):
+            titles = [video.series] + video.alternative_series
+            season = video.season
+            episode = video.episode
+        else:
+            titles = [video.title] + video.alternative_titles
+
+        for title in titles:
```
Diaoul: As I said before, this is going to flood the servers with many useless requests sometimes returning duplicated results. I'm really not a big fan of this.

subliminal/providers/cinemast.py
```diff
@@ -0,0 +1,231 @@
...
+            if subtitles:
+                return subtitles
+
+        return []
+
+    def download_subtitle(self, subtitle):
+        # download
+        url = self.server_url + 'subtitle/download/{}/'.format(subtitle.language.alpha2)
+        params = {
+            'v': subtitle.releases[0],
```
Diaoul: Does the result change if you take another release?

subliminal/providers/cinemast.py
```diff
@@ -0,0 +1,231 @@
...
+                matches.add('series')
+            # season
+            if video.season and self.season == video.season:
+                matches.add('season')
+            # episode
+            if video.episode and self.episode == video.episode:
+                matches.add('episode')
+            # guess
+            for release in self.releases:
+                matches |= guess_matches(video, guessit(release, {'type': 'episode'}))
```
Diaoul: This is potentially going to give an unreal good score to the subtitles as this is the union of all releases instead of picking the best.

subliminal/providers/cinemast.py
```diff
@@ -0,0 +1,231 @@
...
+        # movie
+        elif isinstance(video, Movie):
+            # guess
+            for release in self.releases:
+                matches |= guess_matches(video, guessit(release, {'type': 'movie'}))
```
Diaoul: Same here

subliminal/providers/cinemast.py
```diff
@@ -0,0 +1,231 @@
...
+                    # add the release and increment downloaded count if we already have the subtitle
+                    if subtitle_id in subtitles:
+                        logger.debug('Found additional release %r for subtitle %r', release, subtitle_id)
+                        bisect.insort_left(subtitles[subtitle_id].releases, release)  # deterministic order
+                        subtitles[subtitle_id].downloaded += 1
+                        continue
+
+                    # otherwise create it
+                    subtitle = self.subtitle_class(language, page_link, title, season, episode, title, subtitle_id,
+                                                   subtitle_key, [release])
```
Diaoul: 1 release = 1 instance

subliminal/providers/cinemast.py
ofir123: API is accessible only to registered users.

subliminal/providers/cinemast.py
ofir123: Verification is for user input (in case `None` or an `''` were inserted)

subliminal/providers/cinemast.py
ofir123: Not really..
It's not my site so I'm just guessing and checking how to replace each character.
I'll added dot and semicolons though! Good idea.

subliminal/providers/cinemast.py
ofir123: Done!

subliminal/providers/cinemast.py
ofir123: Done!

subliminal/providers/cinemast.py
ofir123: Done!

subliminal/providers/cinemast.py
Diaoul: Make this a function so it doesn't look too long.

subliminal/providers/cinemast.py
fernandog: @Diaoul It will stop in the first result found. And dogpile caches it so next time it won't hit server.

subliminal/providers/cinemast.py
fernandog: @Diaoul urlencode is from urllib right? Requests already does the encoding and it's not needed. confirm?

subliminal/providers/cinemast.py
ofir123: Not sure I understand what you mean..
If the subtitle matches multiple releases, I want to check all of them, no?

subliminal/providers/cinemast.py
fernandog: do like legendastv:

`if self.username is not None and self.password is not None:`

subliminal/providers/cinemast.py
Diaoul: In that case, don't provide a default user, just use this one for testing and require USER and PASSWORD to be provided in the CLI.
I don't think we have private-only providers so this is all new for you to do.

subliminal/providers/cinemast.py
Diaoul: `self.username = username or self.default_username` this already takes care of that.

subliminal/providers/cinemast.py
Diaoul: For `params` or `data` you are correct but even for the URL requests does that? Anyway, I'm not sure this makes any sense as with special characters the URL will probably be wrong no? If you can add tests that pass with accentuated characters and weird characters that'll be fine for me.

ofir123:
OK, fixed multiple releases issue.

subliminal/providers/cinemast.py
Diaoul: I don't know in that case if we shouldn't do a single Subtitle instance for each release. Even if this is the same actual subtitle.
I think for now if it does the job leave it as is but that's a thought for the long term.
@pannal: what about harmonizing the Subtitle object so that it's not provider specific but rather "subliminal"-specific? This would avoid you patching all the subtitle classes to change the guessing rules. Not sure the feasability though.

subliminal/providers/cinemast.py
ofir123: Done!

subliminal/providers/cinemast.py
ofir123: I really don't think this is necessary for this provider, since it's not really private.
Only the API requires user and password, and the actual website doesn't.
They don't really care about it being a default user as far as I can tell..

Diaoul: If/When they introduce a limited number of downloads per user this is not going to work anymore. Moreover this is not a widely used providers as per the only language it supports. I don't think people would mind entering a user/password in their command line.
Unless you have explicit agreement with Cinemast to use a default user I'd rather require subliminal users to register.

ofir123: OK, removed.

ofir123:
@fernandog How do you fix the tests? Is there something I'm doing wrong with the cassettes?

fernandog:
all good now

ofir123:
@Diaoul @fernandog So what's missing for the merge?

smoresmores:
What is the status of this?

ofir123:
@smoresmores It's ready.. just waiting for the merge..

ofir123:
Hey @ratoaq2 !
Any chance this will get merged in the upcoming version?

PR #823, opened 2017-10-15
Use subliminal to submit full content of folder.

VeggieVampire:
Use subliminal to submit full content of folder.
example 
./batch_subliminal.sh /home/user/media/


Diaoul:
I'm not sure I understand what you're trying to do. If you want subliminal to perform a search on a whole folder you can just pass the folder as argument it will recursively search for subtitles in there.

fernandog:
@VeggieVampire don't need that. just pass folder as argument. Tested again here and works

PR #838, opened 2017-11-25
Minor fix to providers: Subtitle class named arguments

fernandog:
@Diaoul Other providers already have this.


Diaoul:
What does this fix exactly?

fernandog:
@Diaoul to be consistant using named params.

Diaoul:
OK then, LGTM

Issue #768, opened 2017-06-22
How to check if subtitles were saved to disk?


elieobeid7:
How to check if subtitles were successfully saved to disk?

`if (save_subtitles(video, subtitles[video])): `

Or something else?

ratoaq2:
It would really help if the saved subtitle keeps the stored location (path):

Something like this:
```
for subtitle in subtitles:
  if subtitle.path:
    print(subtitle.path)
```

ratoaq2:
Now you can get the subtitle path using the method:

```
subtitle.get_path(self, video, single=False)
```

3e3e410a0132faf17a3ed79b0934196a31a72983


So, you can easily check if the path exists

PR #839, opened 2017-11-26
Store saved subtitle path in Subtitle class after call `save_subtitles`

fernandog:
Related to https://github.com/Diaoul/subliminal/issues/768
@Diaoul what do you think?

ping @ratoaq2


Diaoul:
I don't like it as the Subtitle class is what allows to represent a subtitle from a provider POV. I see no added value in doing this and this adds coupling between the Subtitle class and the saving functions.

fernandog:
Ok thanks

Issue #840, opened 2017-11-26
Missing country data


shlomiLan:
[here ](https://github.com/Diaoul/subliminal/blob/master/subliminal/video.py#L168) you create an `Episode` object, but you don't try to save the country data that is returned by `guessit`. why is that? this information could be helpful when trying to find the correct show in TVDB.

fernandog:
@shlomiLan do you have an idea in mind ? feel free to do a PR to `develop` branch

ratoaq2:
Fixed by 676354ca8b341432e416e42d8d8500a96d0af398

PR #844, opened 2017-11-29
Save country data from GUESSIT result 

shlomiLan:
This PR is for fixing #840 .

### DIFF
--- a/subliminal/refiners/tvdb.py
+++ b/subliminal/refiners/tvdb.py
@@ -264,6 +264,9 @@ def refine(video, **kwargs):
 
     # search the series
     logger.info('Searching series %r', video.series)
+    if video.country:
+        video.series += ' ({})'.format(video.country)
+
     results = search_series(video.series.lower())
     if not results:
         logger.warning('No results for series')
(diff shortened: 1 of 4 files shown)


fernandog:
@shlomiLan thanks. I changed branch to develop. All good now.

What about the TVDB thing you said ?

shlomiLan:
@fernandog Same tests are failed, but I don't think that it because things that I changed, how can I be sure?

fernandog:
you need to review all failing tests

like this error for example:
Now it matches title and imdb_id but it's not expected (at least not in develop code). Now with your PR it seems that is matching that now

>       assert matches == {'series', 'year', 'season', 'episode', 'format', 'video_codec'}
E       AssertionError: assert set(['episode...'title', ...]) == set(['episode'...dec', 'year'])
E         Extra items in the left set:
E         'imdb_id'
E         'title'

fernandog:
@Diaoul what do you think of this? its not wanted/not good he can change strategy

ratoaq2:
Fixed by 676354ca8b341432e416e42d8d8500a96d0af398

Issue #858, opened 2017-12-21
Subliminal cannot handle files with multi episodes with OpenSubtitles provider


h3llrais3r:
When using subliminal in combination with the latest guessit version, there is a problem when the series contains multiple episodes.
Guessit returns a list of episodes, while an int is expected in the ```Video``` class.
Example file: ```Marvels.Agents.of.S.H.I.E.L.D.S05E01-E02.720p.HDTV.x264-AVS.mkv```
```
guessit Marvels.Agents.of.S.H.I.E.L.D.S05E01-E02.720p.HDTV.x264-AVS

For: Marvels.Agents.of.S.H.I.E.L.D.S05E01-E02.720p.HDTV.x264-AVS
GuessIt found: {
    "title": "Marvels Agents of S.H.I.E.L.D.",
    "season": 5,
    "episode": [
        1,
        2
    ],
    "screen_size": "720p",
    "format": "HDTV",
    "video_codec": "h264",
    "release_group": "AVS",
    "type": "episode"
}
```

When searching subtitles with subliminal in combination with opensubtitles provider:

![image](https://user-images.githubusercontent.com/1945295/34260317-bf0b064a-e665-11e7-944c-eb4324a11286.png)

Stack trace:
```
2017-12-21 15:43:32,500 ERROR Unexpected error in provider u'opensubtitles'
Traceback (most recent call last):
  File "C:\Tools\GitHub\Auto-Subliminal\lib\subliminal\core.py", line 134, in list_subtitles_provider
    return self[provider].list_subtitles(video, provider_languages)
  File "C:\Tools\GitHub\Auto-Subliminal\lib\subliminal\providers\opensubtitles.py", line 228, in list_subtitles
    query=query, season=season, episode=episode, tag=os.path.basename(video.name))
  File "C:\Tools\GitHub\Auto-Subliminal\lib\subliminal\providers\opensubtitles.py", line 183, in query
    response = checked(self.server.SearchSubtitles(self.token, criteria))
  File "C:\Tools\GitHub\Auto-Subliminal\lib\subliminal\providers\opensubtitles.py", line 295, in checked
    raise OpenSubtitlesError(response['status'])
OpenSubtitlesError: 408 Invalid parameters; episode is not number
```

Addic7ed does not seem to have a problem with it. I did not test other providers.

fernandog:
@opensubtitles hey, how should we handle this? API expect a number but video file is multi-episode, so the parameter `episode` is a list.

[...]

PR #859, opened 2017-12-22
OpenSubtitles: Add support to multi-episode search in query search

fernandog:
Fixes https://github.com/Diaoul/subliminal/issues/858

hash search is still the best search method but at least we don't end with an expected error (traceback) for user and provider doesn't get discarded
Using suggested fix from @opensubtitles

### DIFF
--- a/subliminal/providers/opensubtitles.py
+++ b/subliminal/providers/opensubtitles.py
@@ -220,7 +220,7 @@ class OpenSubtitlesProvider(Provider):
         if isinstance(video, Episode):
             query = video.series
             season = video.season
-            episode = video.episode
+            episode = min(video.episode) if isinstance(video.episode, list) else video.episode
         else:
             query = video.title
 
--- a/tests/cassettes/opensubtitles/test_multi_episode.yaml
+++ b/tests/cassettes/opensubtitles/test_multi_episode.yaml
@@ -1,0 +1,483 @@
+interactions:
+- request:
+    body: !!python/unicode '<?xml version=''1.0''?>
+
+      <methodCall>
+
+      <methodName>LogIn</methodName>
+
+      <params>
+
+      <param>
+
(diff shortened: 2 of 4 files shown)


fernandog:
@Diaoul what do you think?

h3llrais3r:
Looks good to me.
In most cases, this will probably for pilots with first and second episode.
So looking for subtitles for the first episode will normally be ok. Subtitles for such releases will never be found when searching on the second episode number.

h3llrais3r:
@fernandog @Diaoul 
Just noticed other problems related to multi episode files....
Everywhere we are referencing the `Video.episode` we should check if this is a list or an int.
Now this is not the case and I get a lot of errors when logging a video with multiple episodes...
```
  File "C:\Tools\GitHub\Auto-Subliminal\lib\subliminal\video.py", line 185, in __repr__
    return '<%s [%r, %d, %dx%d]>' % (self.__class__.__name__, self.series, self.year, self.season, self.episode)
TypeError: %d format: a number is required, not list
```
I propose to rollback this PR and do the fix when creating the video from the guessit guess.
Instead of just taking the `guess['episode]` we should take the first if it's a list.
If we do that, we don't need to adapt all over the place that the episode can be a list...
````python
def fromguess(cls, name, guess):
        if guess['type'] != 'episode':
            raise ValueError('The guess must be an episode guess')

        if 'title' not in guess or 'episode' not in guess:
            raise ValueError('Insufficient data to process the guess')

        return cls(name, guess['title'], guess.get('season', 1), guess['episode'], title=guess.get('episode_title'),
                   year=guess.get('year'), format=guess.get('format'), original_series='year' not in guess,
                   release_group=guess.get('release_group'), resolution=guess.get('screen_size'),
                   video_codec=guess.get('video_codec'), audio_codec=guess.get('audio_codec'))
```

Issue #866, opened 2018-01-18
Addic7ed does not find show id for 'The end of the fucking world'


h3llrais3r:
When trying to find subtitles for ```the.end.of.the.fucking.world.s01e04.720p.web.x264-skgtv.mkv``` with addic7ed, the show id could not be found:
```2018-01-18 19:48:04,638 ERROR No show id found for u'the end of the fucking world' ({'country': None, 'year': None})```

Is there a way to bypass the search for showId based on the show name?
Perhaps a custom mapping could be interesting for handling this kind of issues?

The show is listed as ```The end of the f***ing world``` on addic7ed (http://www.addic7ed.com/show/6652), while the files are released as ```The end of the fucking world```.

Any ideas?

**Comment by h3llrais3r** [comment]
Just been looking around in the code a bit and I saw that the call to get the show_id from addic7ed is cached. So I just tried to put the id myself in the cache while initializing my application, which gave a positive result.
However, the be able to match the series title `The End of the F***ing World`, the series regex in addic7ed must be changed (allowing character `*` in the series name).

Original:
```
series_year_re = re.compile(r'^(?P<series>[ \w\'.:(),&!?-]+?)(?: \((?P<year>\d{4})\))?$')
```
Modified:
```
series_year_re = re.compile(r'^(?P<series>[ \w\'.:(),*&!?-]+?)(?: \((?P<year>\d{4})\))?$')
```

fernandog:
@h3llrais3r can you do a PR against develop? Please use a feature branch.

h3llrais3r:
@fernandog Will do the PR for the fix for the regex, but what about the fact that it's never found because the show name doesn't match?

fernandog:
Are you using Medusa right?

h3llrais3r:
@fernandog Thanks, will have a look at it on how to implement.

h3llrais3r:
@fernandog The problem is not completely fixed.
With the PR, it can be matched in subliminal, but we still have the problem that the Addic7ed ID cannot be found when searching with 'the end of the fucking world'.
Since the id cannot be retrieved, no subs can be searched and no matching can be done on series or alternative series name.
In order to fix it, the search for Addic7ed id should also search on alternative series name if the normal series name is not found.
See: https://github.com/Diaoul/subliminal/blob/develop/subliminal/providers/addic7ed.py#L199-L237

fernandog:
Erase your subliminal.dbm in the cache folder

[...]

PR #867, opened 2018-01-19
Addic7ed: Change regex to allow shows with asterisk

fernandog:
Fix https://github.com/Diaoul/subliminal/issues/866

@h3llrais3r


h3llrais3r:
Looks good to me.

Issue #861, opened 2018-01-01
'.WMV' is not a valid video extension


morpheus65535:
Subliminal seems to be case sensitive on Mac OS when it match against it supported file extension? .WMV doesn't work but .wmv work just fine:

    File "/Library/Python/2.7/site-packages/subliminal/core.py", line 374, in scan_video raise ValueError('%r is not a valid video extension' % os.path.splitext(path)[1])

Ref: https://github.com/morpheus65535/bazarr/issues/30

morpheus65535:
Working, thanks!

PR #870, opened 2018-01-23
Use splitext to get extension and compare lowercase

fernandog:
Fixes https://github.com/Diaoul/subliminal/issues/861

@morpheus65535 can you test this please?


morpheus65535:
@stefandevo You can try this? You were the one having an issue with this. Thanks!

morpheus65535:
I'Ve tested it myself and it work. Thanks!

fernandog:
splitext() is specifically written for returning the extension

ratoaq2:
Yes, but that's not needed. The only required change was to add lower before endswith.

Issue #580, opened 2016-02-10
Ignore samples


d-lindahl:
Something I noticed recently is that subliminal downloads subtitles for samples. Does one ever want this? (what surprises me more though is that people upload subs for samples...)

Should be easy to filter out with just a "sample" not in filename-check, if it's a reasonable request. Otherwise just close! I'll survive!
##

Diaoul:
Sample might as well be part of the title. I'm not sure just finding "sample" in the filename is enough.

brp-david:
True enough, didn't think that far. Though atleast excluding the "standard" scene-naming with filename.sample.mkv should take care of most of them. Though the rules only state that the sample should be in a subdirectory called sample. Might just be too much effort for something very little.

brp-david:
Looked through my library (and read some more scene-rules) and the only "sure" way to detect them seems to be that they're located in a subdirectory called sample. The names are all over the place with sample-filename.mkv filename.sample.mkv filename.[sample].mkv

magnub:
Instead of sample filename filtering why not making it possible to filter on the directory name ([Ss]ample) ?

PR #871, opened 2018-01-23
Ignore "Sample" folders (scene) and sample.mkv|avi|etc. files

fernandog:
Fix https://github.com/Diaoul/subliminal/issues/580

@duramato ok with scene rule?


duramato:
Yes.
But keep in mind sample files themselfs aren't just `sample.extension`.

fernandog:
@duramato guessit detects the `sample` in release name ? if yes, which property?

duramato:
Don't think it doesn't.

duramato:
https://github.com/guessit-io/guessit/issues/516

PR #876, opened 2018-01-25
Revert changes related to multi episode video files

h3llrais3r:
Allowing the Video.episode to be a list causes more problems than it fixes.
In case guessit returns a list of episodes, we only take the first (lowest) episode.
Subtitle providers also do not support a list of episodes. A multi episode subitle will always be listed under the lowest episode number.

See discussion in https://github.com/Diaoul/subliminal/pull/859

### DIFF
--- a/subliminal/providers/opensubtitles.py
+++ b/subliminal/providers/opensubtitles.py
@@ -220,7 +220,7 @@ class OpenSubtitlesProvider(Provider):
         if isinstance(video, Episode):
             query = video.series
             season = video.season
-            episode = min(video.episode) if isinstance(video.episode, list) else video.episode
+            episode = video.episode
         else:
             query = video.title
 
--- a/subliminal/providers/tvsubtitles.py
+++ b/subliminal/providers/tvsubtitles.py
@@ -172,9 +172,6 @@ class TVsubtitlesProvider(Provider):
 
         # get the episode ids
         episode_ids = self.get_episode_ids(show_id, season)
-        # Provider doesn't store multi episode information
-        episode = min(episode) if episode and isinstance(episode, list) else episode
-
         if episode not in episode_ids:
             logger.error('Episode %d not found', episode)
             return []
--- a/subliminal/video.py
+++ b/subliminal/video.py
@@ -169,7 +169,11 @@ class Episode(Video):
         if 'title' not in guess or 'episode' not in guess:
             raise ValueError('Insufficient data to process the guess')
 
-        return cls(name, guess['title'], guess.get('season', 1), guess['episode'], title=guess.get('episode_title'),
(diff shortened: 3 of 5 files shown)


fernandog:
@Diaoul how to proceed with this? I recently added support to multi-ep search in opensubtitles

h3llrais3r:
@Diaoul @fernandog 
The problem is that the rest of the code is not compatible with a `Video.episode` as a `list`.
For example, the string representation (https://github.com/Diaoul/subliminal/blob/develop/subliminal/video.py#L181-L185) crashes when the episode is a `list`.

fernandog:
LTV support multi episode. But I think we should be prepared because there is a lot of multi episode releases

h3llrais3r:
I'll leave it up to you to decide, but like it's implemented now, it breaks...
So I think it's either fixing it everywhere or reverting it back to the way it was, and use the min value in case of a list.

fernandog:
This revert is wrong as it's missing other stuff. I will make a new one

h3llrais3r:
@fernandog  Did I forget to revert something?

fernandog:
Cassettes. When revert something just do: git revert commit-hash

h3llrais3r:
Ok, I just thought to leave the tests there and adapted them to be in sync with the reverted code.

fernandog:
@h3llrais3r about this `Allowing the Video.episode to be a list causes more problems than it fixes.` we always allowed to be a list. We use guessit to guess all info

can you send a log with all your "crashes" when using latest develop?

h3llrais3r:
@fernandog I can send some logs, but these will most likely not cover all cases.
If you allow the episode to be a list, you'll need to check everywhere where the episode is used if it doesn't give problems in case it's a list. 
One I know by heart is where the episode is logged. 
```
  File "C:\Tools\GitHub\Auto-Subliminal\lib\subliminal\video.py", line 185, in __repr__
    return '<%s [%r, %d, %dx%d]>' % (self.__class__.__name__, self.series, self.year, self.season, self.episode)
TypeError: %d format: a number is required, not list
```

fernandog:
@Diaoul this seems an issue also in Master. How to deal with it?

h3llrais3r:
@Diaoul @fernandog That's why I'm trying to convince you to revert all multi-ep changes and make sure that we only work with a single episode for now, until a good solution is found for multi-ep.
If we make sure that we always take a single episode from guessit (even when it returns a list) the error for the string representation should be gone. See changes I did in https://github.com/h3llrais3r/subliminal/blob/0d83fd898c5b1a0354043f9abc98cb28d2cd632c/subliminal/video.py#L165-L179

Since there is only 1 provider who is actually supporting multi-ep, I think this is the best way to go.

PR #882, opened 2018-01-30
Revert "TVSubtitles: Fix TypeError: unhashable type: 'list' when mult…

h3llrais3r:
…i-ep"

This reverts commit 5369ec1c1b9534862b76bae996609fefe2c1e778.


h3llrais3r:
In addition to https://github.com/Diaoul/subliminal/pull/881.
I'm also busy with recreating a PR for the error of the show_id.... but these 2 PR's need to be merged first before I can complete it (otherwise I have conflicts with the tests for multi-ep)

fernandog:
you are also reverting the fix for `unhashable type: 'list' `

h3llrais3r:
Yes, because we are reverting all changes made for multi-ep.

fernandog:
let me talk with diaoul first on how to proceed

h3llrais3r:
Ok, but I hope you understand why I want to revert all changes? 
Because currenlty the develop version does not work at all for multi-ep. It gives errors in logging, and matching will never work because the subtitle.episode (int) is compared with video.episode (list) in case of multi-ep.
That's why I prefer to revert all the multi-ep changes, so we can merge the other PR's first. Then we can have a look on how to tackle the multi-ep issue...

PR #885, opened 2018-02-05
Make sure the episode is always a single-ep

h3llrais3r:
Until we find a proper solution for multi-ep, this makes sure that the code does not crash in case of mult-ep.
Since most providers only support single-ep lookup, we take the lowest episode in case of a multi-ep guess result.


fernandog:
@h3llrais3r subtitle class also can have episode as list

h3llrais3r:
@fernandog Fixed for Subtitle.py

fernandog:
@h3llrais3r test as failing for py 3.3 no idea why

h3llrais3r:
@fernandog I don't think it has anything to do with the changes.
Can we trigger the build again?

fernandog:
I already did that but didn't worked. Trying one more time

h3llrais3r:
I have no idea either why it's failing... 
`AttributeError: type object 'test' has no attribute 'install_dists'`

h3llrais3r:
@Diaoul Any idea why the test is failing?

h3llrais3r:
@fernandog Any idea how to proceed here?
Perhaps I can revert the changes again and see if it builds again?

h3llrais3r:
@fernandog @Diaoul 
I found the reason why the tests are failing for Python 3.3.
The difference with the last successful build is the pytest-runner version. The last successfull run was with pytest-runner version 3.0 while all newer builds are running with pytest-runner version 4.0.
I found this issue on github for pytest-runner: https://github.com/pytest-dev/pytest-runner/issues/39

I see 2 solutions:
- Drop python 3.3 support (since latest pytest version does not longer support it either, see https://github.com/pytest-dev/pytest)
- Switch back to pytest-runner < 4.0

What do you suggest? 
Do I include it in this PR?

fernandog:
I say drop 3.3 in this PR

Diaoul:
Agreed.

fernandog:
@ratoaq2 what do you think?

fernandog:
@Diaoul ok! Didn't see you comment. Sorry

h3llrais3r:
@fernandog @Diaoul 
Should I only adapt it in travis.yml?
Or should I also remove it from setup.py?

fernandog:
I would change both

h3llrais3r:
Done.

fernandog:
@h3llrais3r is there any other provider that can be changed regarding show_id ?

h3llrais3r:
@fernandog What do you mean exactly with show_id? Because this fix is about episode_id...

fernandog:
@h3llrais3r 
This PR also changes addic7ed.py and tvsubtitles, about not erroring the show_id in the first title search

h3llrais3r:
@fernandog No I just merged the latest develop branch inside of it to trigger a new build.

fernandog:
@h3llrais3r revert

h3llrais3r:
@fernandog Done.
Or perhaps it's better to do a hard reset and apply the changes again?

h3llrais3r:
Is it ok now? Can you merge?

PR #895, opened 2018-03-10
Correctly enable the Napiprojekt provider

evoL:
Previously the Napiprojekt provider wasn't available in the CLI. This commit adds it to `extensions.py`, so that it's possible to use it.


ratoaq2:
It was disabled because of #536

Issue #887, opened 2018-02-17
Error in omdb refiner with non ascii characters


h3llrais3r:
When scanning a file with name `Un conte de Noël (2008).mkv`, the omdb refiner seems to fail on non ascii characters. It seems to be a problem with dogpile cache.

```
2018-02-17 23:39:12,530 INFO     [SubChecker :: subliminal.core] Refining video with omdb
2018-02-17 23:39:12,532 ERROR    [SubChecker :: subliminal.core] Failed to refine video u'\\\\LAPTOP\\completed\\Un conte de No\xebl (2008).mkv'
2018-02-17 23:39:12,532 DEBUG    [SubChecker :: subliminal.core] Refiner exception:
Traceback (most recent call last):
  File "C:\Tools\GitHub\Auto-Subliminal\lib\subliminal\core.py", line 598, in refine
    refiner_manager[refiner].plugin(video, **kwargs)
  File "C:\Tools\GitHub\Auto-Subliminal\lib\subliminal\refiners\omdb.py", line 155, in refine
    results = search(video.title, 'movie', video.year)
  File "C:\Tools\GitHub\Auto-Subliminal\lib\dogpile\cache\region.py", line 1223, in decorate
    key = key_generator(*arg, **kw)
  File "C:\Tools\GitHub\Auto-Subliminal\lib\dogpile\cache\util.py", line 42, in generate_key
    return namespace + "|" + " ".join(map(to_str, args))
UnicodeEncodeError: 'ascii' codec can't encode character u'\xeb' in position 14: ordinal not in range(128)
```

h3llrais3r:
Same problem will happen on all places where we use `@region.cache_on_arguments` and values can contain non ascii characters. 
I suppose we should convert all unicode params to byte params when use the dogpile cache?
Or perhaps it's better to fix it directly in dogpile cache itself?

fernandog:
@h3llrais3r `all unicode params to byte params when using the dogpile cache` - how would you do that?

h3llrais3r:
@fernandog Perhaps making a custom decorator for the caching that converts the unicode string params (encode with utf-8) to byte string params before doing the caching?
Not sure if this is a good approach and if it's feasible...

h3llrais3r:
I got something working for omdb, but did not test with others.
Idea is to use a custom `function_key_generator`:

```python
def v2b(value):
    """Convert a value to the binary type value."""
    if isinstance(value, six.text_type):
        return value.encode('utf-8')
[...]

PR #897, opened 2018-03-18
Use custom dogpile cache key generator

h3llrais3r:
Dogpile cache cache only accepts ascii strings as values for the default key generator. Passing non ascii characters results in `UnicodeEncodeError`.
By creating a custom key generator we ensure that the value is converted to a valid native string.
On Python 2: unicode strings are converted to byte strings with utf-8 encoding.
On Python 3: byte strings are decoded with utf-8 encoding.
Fixes https://github.com/Diaoul/subliminal/issues/887 and also other `cache_on_arguments` calls where strings are passed with non ascii characters.

### DIFF
--- a/subliminal/cache.py
+++ b/subliminal/cache.py
@@ -15,19 +15,19 @@ EPISODE_EXPIRATION_TIME = datetime.timedelta(days=3).total_seconds()
 REFINER_EXPIRATION_TIME = datetime.timedelta(weeks=1).total_seconds()
 
 
-def _to_native_str(s):
+def _to_native_str(value):
     if six.PY2:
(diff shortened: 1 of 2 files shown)


fernandog:
@ratoaq2 any comment ?

fernandog:
@h3llrais3r is it possible to add any test ?

h3llrais3r:
Will check if I can add one.

h3llrais3r:
@fernandog Test added.
Test contains both original and custom version, so it's also clear what's the difference is between the versions.

ratoaq2:
Thanks

PR #907, opened 2018-03-28
Update README.rst

eran101:
The change is required in order to run the script without "python" before the script file name
For example:
./script.py


ratoaq2:
Thanks

PR #763, opened 2017-06-18
Added support for Wizdom subtitles.

ofir123:

PR #912, opened 2018-04-06
Added wizdom.xyz provider

tnadav:
wizdom.xyz is a new Hebrew subtitles provider meant to be an alternative to the many other Hebrew subtitles site who shut down.
The API provided by the site is really basic and requires imdb id to be present, but that's ok because subliminal has good refiners.


ofir123:
Already open and pending for a long time.. (#763 )

PR #931, opened 2018-07-28
Update README.rst

jovana:
Adding installation instruction for Mac osx


ratoaq2:
I'm not a Mac OSX user, but I believe pip is also available for Mac, so the general pip installation is sufficient.

And the PR should be created against develop branch

PR #948, opened 2019-01-15
Add Subs4Series provider

pkoutsias:


subliminal/providers/subs4series.py
```diff
@@ -0,0 +1,276 @@
...
+        # release_group
+        if (video.release_group and self.version and
+                any(r in sanitize_release_group(self.version)
+                    for r in get_equivalent_release_groups(sanitize_release_group(video.release_group)))):
+            matches.add('release_group')
+        # resolution
+        if video.resolution and self.version and video.resolution in self.version.lower():
+            matches.add('resolution')
+        # format
+        if video.source and self.version and sanitize(video.source) in sanitize(self.version):
```
ratoaq2: I believe you don't need this block, since the next block using guessit will guess the correct source from `self.version`

subliminal/providers/subs4series.py
```diff
@@ -0,0 +1,276 @@
...
+        show_ids = self._get_suggestions(title)
+
+        matched_show_ids = []
+        for show in show_ids:
+            show_id = None
+            show_title = sanitize(show['title'])
+            # attempt with year
+            if not show_id and year:
+                logger.debug('Getting show id with year')
+                show_id = '/'.join(show['link'].rsplit('/', 2)[1:]) if show_title == '%s %d' % (
```
ratoaq2: Just to keep consistency, we should use the new string format style instead of `%` style

subliminal/providers/subs4series.py
```diff
@@ -0,0 +1,276 @@
...
+            if not show_id:
+                logger.debug('Getting show id')
+                show_id = '/'.join(show['link'].rsplit('/', 2)[1:]) if show_title == title_sanitized else None
+
+            if show_id:
+                matched_show_ids.append(show_id)
+
+        return matched_show_ids
+
+    @region.cache_on_arguments(expiration_time=SHOW_EXPIRATION_TIME, to_str=text_type)
```
ratoaq2: Probably you should only cache if it returned results, like here : https://github.com/Diaoul/subliminal/blob/develop/subliminal/providers/legendastv.py#L249

It was an issue in the other provider that because empty results were cached, it wasn't able to download the subtitles when they got uploaded. People had to wait a long time or clear the cache

subliminal/providers/subs4series.py
```diff
@@ -129,9 +129,9 @@ def _get_suggestions(self, title):
 
         """
         # make the search
-        logger.info('Searching show ids with %r', title)
-        r = self.session.get(self.server_url + self.search_url % title, headers={'Referer': self.server_url},
-                             timeout=10)
+        logger.info('Searching show ids with {!r}'.format(title))
```
ratoaq2: For logging, you should just follow the standard logging in python. Do not apply the string formatting. It was correct how it was before

subliminal/providers/subs4series.py
```diff
@@ -0,0 +1,272 @@
...
+            subtitle = self.subtitle_class(language, page_link, show_title, year_num, version, download_link)
+
+            logger.debug('Found subtitle %r', subtitle)
+            subtitles.append(subtitle)
+
+        return subtitles
+
+    def list_subtitles(self, video, languages):
+        # lookup show_id
+        titles = [video.series] + video.alternative_series if isinstance(video, Episode) else []
```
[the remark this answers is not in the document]
pkoutsias: The only reason I used the isinstance check is to get rid of the IDE warnings

subliminal/providers/subs4series.py
```diff
@@ -0,0 +1,272 @@
...
+        subtitles = []
+        # query for subtitles with the show_id
+        for show_id in show_ids:
+            subtitles += [s for s in self.query(show_id, video.series, video.season, video.episode, video.title)
+                          if s.language in languages]
+
+        return subtitles
+
+    def download_subtitle(self, subtitle):
+        if isinstance(subtitle, Subs4SeriesSubtitle):
```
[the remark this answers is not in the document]
pkoutsias: The only reason I used the isinstance check is to get rid of the IDE warnings

subliminal/providers/subs4series.py
```diff
@@ -0,0 +1,272 @@
...
+    for name in archive.namelist():
+        # discard hidden files
+        if os.path.split(name)[-1].startswith('.'):
+            continue
+
+        # discard non-subtitle files
+        if not name.lower().endswith(SUBTITLE_EXTENSIONS):
+            continue
+
+        return archive.read(name)
```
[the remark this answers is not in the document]
pkoutsias: Sometimes an info txt may be contained in the archive, so I select only the subtitle file

pkoutsias:
@ratoaq2 This provider may bundle multiple releases in one archive, but only if the timings are the same. However, the scoring will not be affected because all releases are included in the subtitle's version, so the release matcher will successfully match against all releases.

I have also created a branch in my fork with all 5 providers merged. Should I open a single PR with all the providers, or leave them as-is?

Issue #916, opened 2018-05-02
Podnapisi.NET JSON API


MasterMind2k:
Hi.

We are experimenting with a JSON api, currently we have these four endpoints ready for JSON response:
  * `/subtitles/search/`
  * `/subtitles/search/advanced` (should be the same as above, only HTML differs)
  * `/subtitles/search/<slug-optional>/<moviedb-publish-id>` (same as first, only exact movie search)
  * `/moviedb/search/`

In short, if requesting a GET call without any additional headers, it will return HTML representation. If adding the header "Accept:application/json" it willl return the results in JSON. A list of supported filters can be seen on the endpoint's HTML page itself. For the "basic" search use that "plus" sign right of "active filters" above the search results.

Please give it a test drive and let us know what is missing, compared to `/ppodnapisi/search` endpoint.

Also note, the basic search has a "smart" keyword parser, the old endpoint does not have it.

ratoaq2:
Is there any documentation about these endpoints?

MasterMind2k:
No official yet.

But all filters can be seen through web search and then apply the Accept header.

ratoaq2:
Fixed by #956

Issue #915, opened 2018-04-27
Searching for "Futurama - S00E03" on Podnapisi takes an unusually long time


hachel:
Hello,

I know this isn't a very detailed description of an issue, but that's because I can't quite figure out what's happening.

Check my log file below for that episode. It's a special episode for the show. If I search it manually on podnapisi I get no results.

If I use subliminal like this: `subliminal --debug download --language en "Futurama - S00E03"` once it gets to podnapisi as a provider it seems to scroll all of the available pages and then starts scoring some results for about 15 minutes.
I don't know what it finds and why it takes so long.

Any help?

Thank you

Here is the debug.log: https://paste.drhack.net/?d2453c65376c48c2#WXAU4HXfYnINy7ulIJAG3n9a1n33X4qdjsqNkE0mBzY=

ratoaq2:
Fixed by #956

PR #956, opened 2019-02-24
Switch Podnapisi.NET to json API. 

ratoaq2:
- #916 Switch Podnapisi.NET to json API. 
- #915 Fix to properly search tv-series even if season is 0

It would be nice if some users could try it
@MasterMind2k @hachel

### DIFF
--- a/subliminal/providers/podnapisi.py
+++ b/subliminal/providers/podnapisi.py
@@ -1,17 +1,10 @@
 # -*- coding: utf-8 -*-
 import io
+import json
 import logging
-import re
 
 from babelfish import Language, language_converters
 from guessit import guessit
-try:
-    from lxml import etree
-except ImportError:
-    try:
-        import xml.etree.cElementTree as etree
-    except ImportError:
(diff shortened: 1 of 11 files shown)


subliminal/providers/podnapisi.py
```diff
@@ -74,57 +67,56 @@ def __init__(self):
...
-                title = subtitle_xml.find('title').text
-                season = int(subtitle_xml.find('tvSeason').text)
-                episode = int(subtitle_xml.find('tvEpisode').text)
-                year = int(subtitle_xml.find('year').text)
+                language = Language.fromietf(data['language'])
+                hearing_impaired = 'hearing_impaired' in data['flags']
+                page_link = data['url']
+                releases = data['releases'] + data['custom_releases']
+                title = data['movie']['title']
+                season = int(data['movie']['episode_info']['season']) if is_episode else None
```
MasterMind2k: Season and episode may not be defined. I believe using a `get` with apropriate default value would be a better choice.

PR #969, opened 2019-05-26
Fix (rename) file using release name

andreoliwa:
First of all, thanks for subliminal, it's the best subtitle downloaded I tried so far.

This is a hacky solution to solve a problem I had.
I recovered some movies from a failed hard drive, but the file names were gone.

So I modified subliminal to download subtitles by the file hash, then rename the movie file according to the release name.
I added an option `--fix-filename` to the `download` command.

This solution has several problems and hardcoded behaviours.

1. It only works with OpenSubtitles so far;
2. It breaks with some providers;
3. It loops on other providers;
4. The first release name is taken by default;
5. A subdirectory is created if different from the release name;
6. It only works with movies, not episodes;
7. Tests are broken.

If someone else is interested in this feature, then we can all brainstorm and suggest a better user interface for it.
Examples of some possibilities:

1. A separate command `rename` instead of `download`, with some common options and arguments;
2. Providers could have a boolean `allow_hash_search` attribute to indicate if the provider can search movies/episodes by hash;
3. Only accept the `--provider` option for providers that allow searching by hash;
4. Option `--use-release=ask/first/random`, if a movie has more than one release name matching the hash;
5. Use some other attribute to rename the file if the release name is not available.

I hope the feature is useful for someone else, even in the current raw state it's in.
Any suggestion, comment or correction is welcome.

### DIFF
--- a/subliminal/cli.py
+++ b/subliminal/cli.py
@@ -320,10 +320,12 @@ def cache(ctx, clear_subliminal):
 @click.option('-z/-Z', '--archives/--no-archives', default=True, show_default=True, help='Scan archives for videos '
               '(supported extensions: %s).' % ', '.join(ARCHIVE_EXTENSIONS))
 @click.option('-v', '--verbose', count=True, help='Increase verbosity.')
+@click.option('-ff', '--fix-filename', is_flag=True, default=False, help='Fix video file name according to the release '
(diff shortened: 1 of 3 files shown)


getzze:
I'm going to close this as it looks more like a way to use `subliminal` to fit your needs than a feature that other people would use.

Issue #925, opened 2018-06-15
Feature request: parameter to choose between normal and forced subtitles


morpheus65535:
Would it be possible to have a parameter in download_best_subtitles() to return only forced subtitles? You know those that only appear when they speak in a foreign language during a movie or episode?

It's been asked by users there:
https://github.com/morpheus65535/bazarr/issues/88
https://github.com/morpheus65535/bazarr/issues/89

Thanks!

jackjohnsonuk:
Yes, I'd love this too please!
Thanks in advance

goose-ws:
+1 for would really appreciate this feature.

I don't need subtitles for the entire Game Of Thrones episodes, just the parts in Dothraki / Valyrian / other non-english languages.

JakeShirley:
I would also like this.  If this is supposed to be a legit acquisition platform this will be necessary.

spaceboy76:
a setting for forced subs would make this complete

PieMistake:
Looks like someone has written a patch to support it, but it’s failed. 

Any chance someone could take a look? It would be great if we could get support for foreign part (forced) subtitles only.

timcoop3r:
+1
With Plex killing off plugins and no more updates for SubZero I was hoping subliminal supported this so I could use FlexGet to pull in subs instead. I suspect more and more people will now be having these issues unless Plex implements something themselves for Forced subs

morpheus65535:
@timcoop3r if you are using Sonarr or Radarr, you can take a look at Bazarr.

JakeShirley:
@morpheus65535 I tried using Bazarr and ended up here.  The dev won't implement forced subtitles because it uses this package (subliminal) to fetch subtitles and subliminal doesn't support forced subtitles only.

morpheus65535:
@JakeShirley I’m Bazarr main dev and it support forced subtitles since v0.8.0.

getzze:
completed in #1175

PR #977, opened 2019-07-17
Implements foreign parts only support with OpenSubtitles

JonBons:
Just a quick attempt at seeing how difficult it would be to support downloading foreign parts only subtitle files with OpenSubtitles.

Implements https://github.com/Diaoul/subliminal/issues/925 and I'm not a huge python guy so I'm sure I've done some things wrong here, I guess the biggest problem is the changes to parameter order with some functions and not updating the tests.


ratoaq2:
You need to create a PR against develop branch (not master)

PR #998, opened 2020-04-16
do not crash when Subtitle.guess_encoding fails

forresty:
Sometimes `Subtitle.guess_encoding` will fail and return `None` therefore result in the CLI to crash.

Specifically, `chardet.detect` might return `None`.

This tiny patch will avoid it.

A better approach might be improving `guess_encoding`, but I am not confident enough to do that (at least for now)

thanks.

### DIFF
--- a/subliminal/subtitle.py
+++ b/subliminal/subtitle.py
@@ -73,7 +73,11 @@ class Subtitle(object):
         if self.encoding:
             return self.content.decode(self.encoding, errors='replace')
 
-        return self.content.decode(self.guess_encoding(), errors='replace')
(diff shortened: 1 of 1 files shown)


ratoaq2:
I strongly believe that this is already fixed on the develop branch. A PR should be created against develop branch

forresty:
@ratoaq2 thanks, confirmed it's fixed on develop branch, I will use that branch for now.

Any plan on merging dev branch to master and create a new release? seems dev branch is 100+ commits ahead now.

ratoaq2:
I don't have account or permission to create a new release. I only have permission to commit in this github repo

Issue #974, opened 2019-07-01
TypeError: expected string or bytes-like object


nicolabeghin:
Hi everyone
I'm currently experiencing an issue with subliminal used through official Docker image. I already made sure guessit is updated as in #894 and #775. 

> subliminal, version 2.1.0.dev
> guessit, version 3.0.4

    Traceback (most recent call last):
      File "/usr/local/bin/subliminal", line 11, in <module>
        load_entry_point('subliminal', 'console_scripts', 'subliminal')()
      File "/usr/local/lib/python3.7/site-packages/click/core.py", line 764, in __call__
        return self.main(*args, **kwargs)
      File "/usr/local/lib/python3.7/site-packages/click/core.py", line 717, in main
        rv = self.invoke(ctx)
      File "/usr/local/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
        return _process_result(sub_ctx.command.invoke(sub_ctx))
      File "/usr/local/lib/python3.7/site-packages/click/core.py", line 956, in invoke
        return ctx.invoke(self.callback, **ctx.params)
      File "/usr/local/lib/python3.7/site-packages/click/core.py", line 555, in invoke
        return callback(*args, **kwargs)
      File "/usr/local/lib/python3.7/site-packages/click/decorators.py", line 27, in new_func
        return f(get_current_context().obj, *args, **kwargs)
      File "/usr/src/app/subliminal/cli.py", line 436, in download
        hearing_impaired=hearing_impaired, only_one=single)
      File "/usr/src/app/subliminal/core.py", line 260, in download_best_subtitles
        for s in subtitles], key=operator.itemgetter(1), reverse=True)
      File "/usr/src/app/subliminal/core.py", line 260, in <listcomp>
        for s in subtitles], key=operator.itemgetter(1), reverse=True)
      File "/usr/src/app/subliminal/score.py", line 106, in compute_score
        matches = subtitle.get_matches(video)
      File "/usr/src/app/subliminal/providers/podnapisi.py", line 53, in get_matches
        sanitize(name) for name in [video.series] + video.alternative_series)):
      File "/usr/src/app/subliminal/providers/podnapisi.py", line 53, in <genexpr>
        sanitize(name) for name in [video.series] + video.alternative_series)):
      File "/usr/src/app/subliminal/utils.py", line 111, in sanitize
        string = re.sub(r'[%s]' % re.escape(''.join(characters)), ' ', string)
[...]

PR #1000, opened 2020-04-21
Guessit sometimes returns a list as the series

guss77:
This fixes issue #974 for me.


ratoaq2:
This is already fixed in develop branch

PR #1021, opened 2020-12-21
Add offset option for download command

### LINKED ISSUE #1018: A method to download an alternate sub if the intially downloaded sub is no good
Sometimes subliminal will download subs that are bad, for instance I have this movie, white snake, and the downloaded sub for it is 
_White.Snake.2019.1080p.BluRay.x264.AAC5.1-[YTS.LT].en.srt_

It's (entire) content:

```
`1
00:00:30,000 --> 00:00:36,000
Provided by YTS.LT

2
00:00:36,500 --> 00:00:42,000
Find the official YIFY movies site at
https://YTS.LT

3
00:30:30,000 --> 00:30:36,000
Downloaded from YTS.LT

4
01:00:30,000 --> 01:00:36,000
Download more movies for free
from YTS.LT
`
```

Nevermind being a mistimed subtitle file which can sometimes happen, it's just not a real subtitle at all.

I'd like to have the ability to make subliminal look for an alternative subtitle file instead of the one it first downloaded for a file.

Additionally perhaps an automated feature that checks if the filesize of the subtitle file targeted for download is at least over 1kb and rejects the file if it isn't to prevent this specific scenario might be nice.

Omrigan:
Closes #1018

### DIFF
--- a/subliminal/cli.py
+++ b/subliminal/cli.py
@@ -319,11 +319,13 @@ def cache(ctx, clear_subliminal):
 @click.option('-w', '--max-workers', type=click.IntRange(1, 50), default=None, help='Maximum number of threads to use.')
 @click.option('-z/-Z', '--archives/--no-archives', default=True, show_default=True, help='Scan archives for videos '
               '(supported extensions: %s).' % ', '.join(ARCHIVE_EXTENSIONS))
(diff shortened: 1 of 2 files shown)


getzze:
Thanks for contributing but we will go with a cli option to skip subtitles by id instead, because it's more versatile.

PR #1028, opened 2021-04-06
Limit versions of rarfile that can be used

gene1wood:
In markokr/rarfile#70 it's reported that the `custom_check` function was removed from the rarfile pypi package. As a result the error reported in Diaoul/subliminal#1013 occurs when a version of rarfile newer than 3.1 is installed.

This fixes #1013 by limiting the range of rarfile versions.

Once the legendastv provider is changed to no longer use `custom_check` this version constraint could be removed.

PR #1045, opened 2021-12-24
Fix removal of custom check from rarfile dependency

eSoares:


clach04:
setup.py should also be updated to require `>3.1` - see #1028 1028

getzze:
legendastv Provider was removed

PR #1067, opened 2024-04-30
Add Provider for opensubtitles.com

### LINKED ISSUE #1063: opensubtitle API deprecation, New REST API and VIP account
The old opensubtitle.org API has been deprecated and only will work in the future for VIP accounts.
With these changes, the method used by subliminal is no longer working and only download a subtitle with an ad.
The info at this time is a bit confusing and trends to imply that only VIPs will can download subtitles with the API.
But it seems to be only true for the old API from dot org.
For example, I found this project https://github.com/dusking/opensubtitles-com that use the new APi from dot com and it can download subtitles only with the user and password without pay.

Screenshot_20240105_054845

Is there any chance that it will be used in subliminal in the future?

getzze:
Closes #1063

Please test with the `develop` branch.


subliminal/extensions.py
```diff
@@ -100,7 +102,7 @@ def unregister(self, entry_point):
 ])
 
 #: Disabled providers
-disabled_providers = ['addic7ed', 'napiprojekt', 'opensubtitlesvip', 'shooter']
+disabled_providers = ['addic7ed', 'napiprojekt', 'opensubtitles', 'opensubtitlesvip', 'shooter']
```
ptrcnull, line 105: can we wait a little bit with this one? ( #1068 )

getzze:
I had understood from the discussion in #1063 that the old API was not available anymore. It's great if it can be repaired.

I will review #1068 when I have time, thanks!

RaXorX:
Tested https://github.com/Diaoul/subliminal/pull/1068 which worked totally fine.

subliminal/extensions.py
getzze: kept the old API provider

subliminal/providers/opensubtitlescom.py
```diff
@@ -0,0 +1,626 @@
...
+        if tmdb_id:
+            criterion.update({'tmdb_id': str(tmdb_id).lstrip("0")})
+            if show_tmdb_id:
+                criterion.update({'parent_tmdb_id': str(show_tmdb_id).lstrip("0")})
+
+        if opensubtitles_id:
+            criterion.update({'id': opensubtitles_id})
+
+        if query:
+            criterion.update({'query': query.replace('\'', '')})
```
ptrcnull, line 350:
```suggestion
            criterion.update({'query': query.replace("'", '')})
```

subliminal/converters/opensubtitlescom.py
```diff
@@ -0,0 +1,32 @@
+# -*- coding: utf-8 -*-
```
ptrcnull, line 1: AFAIK this is the default with modern Python versions

getzze: Yes, we can remove these lines from all the files as part of #1071

ptrcnull (approved):

ptrcnull:
personal nitpick: squashing the commits, leaving only `add Provider for OpenSubtitles.Com` and `add api key to omdb and clean tvdb`

getzze:
Thanks for reviewing!
Note for the future, `CachedSession` from `requests_cache` is used as a default (call `OpenSubtitlesComProvider(cached=False)` to disable) but it is not tested as it clashes with `vcr`.

PR #1068, opened 2024-04-30
Fix old OpenSubtitles provider

ptrcnull:


RaXorX:
This works just fine. I can't root for the VIP portion if com has any subscription. But the normal one works totally fine.

getzze (approved):

PR #1069, opened 2024-05-06
Add Gestdown Provider

getzze:
Addic7ed proxy

### DIFF
--- a/subliminal/providers/gestdown.py
+++ b/subliminal/providers/gestdown.py
@@ -1,0 +1,298 @@
+# -*- coding: utf-8 -*-
+import logging
+import re
+
+from babelfish import Language
+from guessit import guessit
+from requests import HTTPError, Session
+
+from . import Provider
+from ..exceptions import DownloadLimitExceeded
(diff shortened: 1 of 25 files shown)


subliminal/providers/gestdown.py
```diff
@@ -0,0 +1,298 @@
...
+        """Search the show id from the `series`.
+
+        :param str series: series of the episode.
+        :param int series_tvdb_id: tvdb id of the series.
+        :return: the show id, if found.
+        :rtype: int
+
+        """
+        if self.session is None:
+            return
```
ptrcnull, line 117: maybe we could raise an exception when the provider isn't initialized but is attempted to be used?

getzze: That would be better, you are right!
Although that is a problem that is present in all the providers, so it would be better to create a specific `Exception` for it and correct all the provider. I think this is out of scope of this PR.

getzze: Ah yes, can you open an issue please?

PR #1072, opened 2024-05-07
Drop old python versions (<3.8)

getzze:
In this PR:
- [x] removed the `# -*- coding: utf-8 -*-` line
- [x] remove the `six` reference and requirement
- [x] remove the `sys.version` check for python version older than 3.8
- [x] remove  `--pep8 --flakes` options from `pytest` addopts
- [x] remove subclassing from `object`
- [x] remove full usage of `super()` 
- [x] remove `pkg_resources` dependency
- [x] remove call to `datetime.utcnow()`
- [x] remove escaped `**` from docstrings
- [x] add a `NotInitializedProviderError` Exception (#1070)
- [x] add `from __future__ import annotations` line for easier type annotations
- [x] update dependencies [EDITED]

Partially solves #1071

Issue #1071, opened 2024-05-06
Drop support for python 2 and python < 3.8


getzze:
Dropping support for old python versions will help to clean the code and to add type annotations.

Tasks:
- [x] remove `six` from dependencies ( and replace `six.text_type` with `str`, etc. ) (#1072)
- [x] remove the `-*- coding: utf-8 -*-` lines  (#1072)
- [x] add a `pyproject.toml` file.  (#1073)
- [x] update outdated dependencies  (#1072)
- [x] repair the tests (#1074)

ptrcnull:
this might include:
- removing `six` from dependencies ( and replacing `six.text_type` with `str`, etc. )
- removing the `-*- coding: utf-8 -*-` lines

getzze:
also using a `pyproject.toml` file with all the relevant parameter for installing, testing and linting the files.

getzze:
Done!

PR #1073, opened 2024-05-08
Use pyproject.toml

getzze:
In this PR:
- [x] create a `pyproject.toml`
- [x] remove the `setup.{cfg,py}`, `{dev-,}requirements.txt` and `pytest.ini` files.
- [x] add configuration for using `ruff`, `mypy` and `typos`.
- [x] create a separate `_version.py` file that holds only the `__version__` variable. For later use of `bump-my-version` for instance.

Needs to be merged after #1072 
Partially solves #1071


pyproject.toml
```diff
@@ -0,0 +1,223 @@
...
+build-backend = "setuptools.build_meta"
+
+# https://peps.python.org/pep-0621/
+[project]
+name = "subliminal"
+description = "Subtitles, faster than your thoughts"
+requires-python = ">=3.8"
+license = { text = "MIT" }
+authors = [{ name = "Antoine Bertin", email = "diaoulael@gmail.com" }]
+maintainers = [{ name = "Antoine Bertin", email = "diaoulael@gmail.com" }]
```
Diaoul, line 13: Please add yourself and other maintainers :pray:

ptrcnull:
could we perhaps use something like `setuptools_scm`? i've found it the most convenient so far, both for maintenance as a project developer and also packaging for Linux distro usage

pyproject.toml
```diff
@@ -0,0 +1,223 @@
...
+    "I",    # isort
+    "UP",   # pyupgrade
+    "S",    # bandit
+    "C4",   # flake8-comprehensions
+    "B",    # flake8-bugbear
+    "A001", # flake8-builtins
+    "TCH",  # flake8-typecheck
+    "TID",  # flake8-tidy-imports
+    "RUF",  # ruff-specific rules
+]
```
Diaoul, line 143: Is this necessary or can we go with the default?

pyproject.toml
```diff
@@ -0,0 +1,223 @@
...
+
+
+# https://docs.astral.sh/ruff/
+[tool.ruff]
+line-length = 120
+src = ["subliminal", "tests"]
+
+[tool.ruff.lint]
+pydocstyle = { convention = "pep257" }
+select = [
```
ptrcnull, line 128: i'd also add `ISC` and perhaps `PT`

Diaoul (approved):
Massive overhaul, thanks!
LGTM

pyproject.toml
ptrcnull: default ruff config doesn't include much, unfortunately

Diaoul: Ah bummer and very counterintuitive :disappointed:

getzze:
> could we perhaps use something like `setuptools_scm`? i've found it the most convenient so far, both for maintenance as a project developer and also packaging for Linux distro usage

I used it before but it has a huge problem: it needs all the .git folder to install. So you cannot install from the source tarball (which is the main install media for Arch Linux for instance).
I personnally switched to `bump-my-version` (https://github.com/callowayproject/bump-my-version/) after realizing this.

But maybe I can revert creating a `_version.py` file and we decide later.

pyproject.toml
getzze: Done!
Note: ISC001 conflicts with `ruff-format`, we can exclude it later if it prevents automatic linting

getzze:
I reverted creating `_version.py` and created an issue to discuss solutions: #1076

pyproject.toml
getzze: One alternative (that I use in personal project) is to put `select = ["ALL"]` and ignore specific rules.
Like that, if a new version of ruff adds new rules, they are automatically included.
But then we would need to pin ruff version to avoid adding rules automatically...

getzze:
wait for #1082

getzze:
good for a final review :)

pyproject.toml
ptrcnull: i don't think we'd want to do `select = ["ALL"]` ( e.g. I use my system `ruff` install which could be updated separately from subliminal, so that would make it fail for just me but not others )

getzze: I agree

pyproject.toml
getzze: maybe some of `ANN` would be useful

getzze:
I added some rules for ruff: FA, BLE, RET, SIM, DTZ, ASYNC, TRY and ANN2xx and ANN001 (the other are painful)

One doubt that I have, should we add this? :
```
[tool.ruff.format]
quote-style = "single"
```
I prefer double-quote, but given that historically, all the strings in subliminal are single-quote, it will be a huge change... Done automatically by `ruff` but still.

ptrcnull:
changing all quotes would be a mess in git history, imo we can keep the current style ( and enforce it with lints ) rather than change it

getzze:
added the `quote-style = single` option

getzze:
Anything else or do you thing this can be merged @ptrcnull ?

ptrcnull:
i think the only thing left is to add the current maintainers

getzze:
hey @ptrcnull , can you add yourself to the list of maintainer so we can merge this PR please?

getzze:
Thanks, merged!

PR #1074, opened 2024-05-09
Repair tests (without recomputing most of the cassettes)

getzze:
Mostly changes to the `tests` files. The changes to the `subliminal` module are minimal.


getzze:
If we delete the old cassettes then more tests are failing, mainly due to Providers that stopped working (thesubdb, addic7ed, ...). These are for other PRs.

getzze:
Ok, this is good now (I had to remove some python2 code also).

After this PR, only the `refiners` and `providers` tests are failing.
@ptrcnull if you want to give a quick look, so we can merge this before #1073

ptrcnull (approved):

PR #1077, opened 2024-05-10
Opensubtitles.com cli options and document (also Gestdown)

getzze:
Simple PR:
- document OpenSubtitlesCom and Gestdown Providers
- add them to the entry points
- add the option to pass credentials for Opensubtitles.com


subliminal/refiners/hash.py
```diff
@@ -10,6 +10,8 @@
     'napiprojekt': hash_napiprojekt,
     'opensubtitles': hash_opensubtitles,
     'opensubtitlesvip': hash_opensubtitles,
+    'opensubtitlescom': hash_opensubtitles,
```
ptrcnull, line 13: are we sure they're using the same hashing algorithm?

getzze: I think so:
https://opensubtitles.stoplight.io/docs/opensubtitles-api/e3750fd63a100-getting-started#calculating-moviehash-of-video-file

getzze:
I will merge this, as the changes are not very big, it is just to make public the 2 new providers. So people can test them.

PR #1081, opened 2024-05-13
Fix addic7ed

getzze:
Fixed the `get_show_ids` function.

Used list of user-agents from Bazarr.


subliminal/providers/_agent_list.py
```diff
@@ -0,0 +1,783 @@
+FIRST_THOUSAND_OR_SO_USER_AGENTS = [
```
ptrcnull, line 1: i'd just get a really generic Chrome on Windows User-Agent, can't see a reason to have *this many* of them

getzze: I put  a list of 6 user-agents in the `addic7ed.py` file.

getzze:
I added type hints.

ptrcnull (approved):
to be fair, some of these MRs are so big, reading the diff doesn't even make sense - i have to open the new code in an IDE and read through it :p

slightly grumpy at the massive diffs from test cassettes, but there's not much we can do about it if we want to keep tests working properly 😅

getzze:
Yep, sorry...
Actually I don't expect you to review them, I'll ping you when there is a PR with real changes.

PR #1089, opened 2024-05-28
Update docker file to 3.9.19-alpine3.20

DerouineauNicolas:
Hi,
In order to test the new branch, I find it easier to use docker.

I updated the one I found with unrar installation. I tested with:

    docker build -t test .
    docker run test --debug --opensubtitlescom myuser mypassword download -p opensubtitlescom -sfl en "The Matrix"

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,25 +2,24 @@ FROM python:3.9.19-alpine3.20
 
 MAINTAINER Antoine Bertin <diaoulael@gmail.com>
 
-RUN apk add curl
-
 # set version label
+ARG BUILD_WITH_UNRAR=false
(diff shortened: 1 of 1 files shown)


getzze:
`unrar`, through the `rarfile` python package, is used to read video RAR archives.

It could be made an optional dependency though, as this use case seems really uncommon.

Or we could use `unar` in the Docker env instead. But I would go for making it an optional dependency.

DerouineauNicolas:
I made the build optionnal. Tested with:

 docker build --build-arg BUILD_WITH_UNRAR=true --no-cache -t test .
 docker build  --no-cache -t test .

getzze:
Thanks, I don't use Docker but it looks good to me.

PR #1092, opened 2024-05-29
Add type hint to subtitle.py

getzze:
Breaking changes:
- `Subtitle` takes one positional argument (`language`), the rest are keyword only.
- ~~`Subtitle.content` is always a `bytes`, never `None`.~~  `content` can be `bytes` or `None`, it will be improved in another PR.
- `Subtitle.text` is always a `str` never a `bytes` or `None`. If the encoding cannot be guessed, return an empty string

### DIFF
--- a/subliminal/subtitle.py
+++ b/subliminal/subtitle.py
@@ -143,6 +143,8 @@ class Subtitle:
         :rtype: str
 
         """
+        if not isinstance(self.content, bytes):
+            return None
         logger.info('Guessing encoding for language %s', self.language)
 
         # always try utf-8 first
--- a/subliminal/subtitle.py
+++ b/subliminal/subtitle.py
@@ -1,11 +1,19 @@
+"""Subtitle class."""
+
 from __future__ import annotations
 
 import codecs
 import logging
 import os
+from typing import TYPE_CHECKING, ClassVar
 
 import chardet
-import srt
+import srt  # type: ignore[import-untyped]
+
+if TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


subliminal/subtitle.py
```diff
@@ -52,38 +83,38 @@ def __init__(self, language, hearing_impaired=False, page_link=None, encoding=No
                 logger.debug('Unsupported encoding %s', encoding)
 
     @property
-    def id(self):
-        """Unique identifier of the subtitle"""
-        return ""
+    def id(self) -> str:
+        """Unique identifier of the subtitle."""
```
ptrcnull, line 87: aah, i just noticed - do we have a linter rule that enforces these dots? not sure if i like them

getzze: it's [D400](https://docs.astral.sh/ruff/rules/ends-in-period/).
it's automatically fixed by ruff.

getzze:
I pushed some changes to revert forcing `content` to be `bytes` and never `None` as it was breaking other stuff.

It should be dealt in another PR, with other improvements. Like, I think decoding the subtitle content every time the `text` property is read should be changed and `text` somehow cached.

ptrcnull:
cleaned up the commits a little ( `Add type hint to subtitle.py` was mixed with `Detect empty subtitle content` and both had related changes ); hopefully it doesn't cause any rebase failures, as the contents are exactly the same

ptrcnull:
ah, but github didn't register the push before i clicked "merge"...

PR #1093, opened 2024-05-29
Add TMDB refiner

getzze:
It is used in some providers from `bazarr`, better make it a proper refiner.

Needs https://github.com/Diaoul/subliminal/pull/1091 to be merged before (it defines `Video.tmdb_id`)

### DIFF
--- a/subliminal/refiners/tmdb.py
+++ b/subliminal/refiners/tmdb.py
@@ -1,0 +1,339 @@
+"""Refine the :class:`~subliminal.video.Video` object by searching on TMDB."""
+
+from __future__ import annotations
+
+import logging
+import re
+from typing import Any, ClassVar, cast
+
+import requests
+
+from subliminal import __short_version__
+from subliminal.cache import REFINER_EXPIRATION_TIME, region
+from subliminal.utils import decorate_imdb_id, sanitize, sanitize_id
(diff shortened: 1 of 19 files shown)


subliminal/refiners/tmdb.py
```diff
@@ -0,0 +1,335 @@
...
+from subliminal.cache import REFINER_EXPIRATION_TIME, region
+from subliminal.utils import sanitize
+from subliminal.video import Episode, Movie, Video
+
+logger = logging.getLogger(__name__)
+
+series_re = re.compile(r'^(?P<series>.*?)(?: \((?:(?P<year>\d{4})|(?P<country>[A-Z]{2}))\))?$')
+
+#: TMDB subliminal API key
+TMDB_API_KEY = '3dac925d5d494853ea6ef9161011fbb3'
```
ptrcnull, line 21: what's the quota on this API key?

getzze: I understood it doesn't have quota: https://developer.themoviedb.org/docs/rate-limiting

But it's a good point in general, should we provide a default API key. It also applies to the OMDB and TVDB refiners.

getzze:
I decided to remove the API key, because it seems more like a personal key than an application key.

It's still used in the tests

PR #1103, opened 2024-06-01
Cache decoding of subtitle content and work with non-srt subtitles

getzze:
Requires #1092 

The `text` property is computed only once when the `content` is changed.
Added more encoding guesses according to the provided language.

Also `is_valid` was only working for `srt` subtitles, which is not compatible with `napiprojekt` for instance.
I added a new dependency to `pysub2` to handle other subtitle format.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,6 +42,7 @@ dependencies = [
     "enzyme>=0.4.1",
     "guessit>=3.0.0",
     "platformdirs>=4.2",
+    "pysubs2>=1.7",
     "rarfile>=2.7",
     "rebulk>=3.0",
     "requests>=2.0",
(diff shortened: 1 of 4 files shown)


ptrcnull:
maybe we could switch to `pysubs2` altogether? we're currently using `srt` ( switched from `pysrt` as well ) to just reparse subtitles ( and i'm not exactly sure what for, if it's invalid srt it would also fail at parsing here..? )

getzze:
For making sure the subs are valid I would say that `pysubs2` is enough. But `srt` seems to be very good at repairing `srt`. Maybe it's useful to add an option to automatically repair `srt`. We open an issue to know if it would be a useful feature?

ptrcnull:
i'm not sure to which extent we can "repair" subs - `srt` was used pretty much explicitly as a validator, would require writing some test cases to see how far we can take the repairing feature

PR #1104, opened 2024-06-01
Fix napiprojekt

getzze:
Requires https://github.com/Diaoul/subliminal/pull/1103

The provider was completely broken: now the raw content is correctly downloaded and following #1103, the subtitles are correctly decoded and validated.

Also add type hints

PR #1105, opened 2024-06-03
Guess subtitle format using pysubs2

getzze:
Correct #1103  to fix #1104

### DIFF
--- a/subliminal/subtitle.py
+++ b/subliminal/subtitle.py
@@ -9,7 +9,7 @@ from typing import TYPE_CHECKING, ClassVar
 
 import chardet
 import srt  # type: ignore[import-untyped]
-from pysubs2 import SSAFile  # type: ignore[import-untyped]
+from pysubs2 import SSAFile, UnknownFPSError  # type: ignore[import-untyped]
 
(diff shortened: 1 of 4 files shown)


subliminal/subtitle.py
```diff
@@ -334,17 +349,20 @@ def guess_encoding(self) -> str | None:
...
         :type video: :class:`~subliminal.video.Video`
         :param bool single: save a single subtitle, default is to save one subtitle per language.
+        :param (str | None) extension: the subtitle extension, default is to match to the subtitle format.
         :return: path of the subtitle.
         :rtype: str
 
         """
-        return get_subtitle_path(video.name, None if single else self.language)
+        if extension is None:
+            extension = FORMAT_TO_EXTENSION.get(self.subtitle_format, '.txt')
```
ptrcnull, line 364: IMO defaulting to `.srt` would be better here, that's what subliminal has been doing all along

getzze: Right, it's better to keep the old behavior.
Anyway the fallback should happen only if the `subtitle_format` was not defined and could not be guessed (meaning it's an invalid format).

PR #1113, opened 2024-06-05
Fix Windows-1250 decoding

### LINKED ISSUE #1112: Fixup NapiProjekt subtitles encoding
sometimes the subtitles are getting returned as Windows-1250 instead of UTF-8; detect those cases and automatically convert them to UTF-8 whenever possible

marking as enhancement because it's NapiProjekt who's being silly and marking non-Unicode subtitles as UTF-8

ptrcnull:
closes #1112


getzze:
Nice trick for checking the encoding, but I think we should also reorder the encoding to put the previous ones first, just to be on the safe side and not change old behaviors.

ptrcnull:
sure, i'll reorder them to how they were before whenever i can

getzze:
I can do it if you want

ptrcnull:
alright, in this PR or in a separate one?

getzze:
I'll commit here if you don't mind

ptrcnull:
thanks!

PR #1117, opened 2024-06-07
Update MANIFEST.in

getzze:
Include `docs` and `tests` folders (except `data` and `cassettes` subfolders).


getzze:
@ptrcnull my only doubt in this PR is if we exclude the `cassettes` and `data` test subfolders from the sdist. This PR excludes them.

ptrcnull:
oh, i'd rather not do that, some Linux distributions pull the sdist tarballs to build the package and they want to run the tests

getzze:
Thanks, I'll change that and merge it

PR #1118, opened 2024-06-07
Add github actions

getzze:
I wrote 2 files:
- one for lint and type checking
- one for the test and coverage. Tests are failing right now.

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -85,27 +85,46 @@ jobs:
           python -m pip install --upgrade pip
           pip install -e .[test]
       - name: run pytest --cov
+        env:
+          COVERAGE_FILE: ".coverage.${{ matrix.os }}.${{ matrix.python-version }}"
+          # Alternatively you can run coverage with the --parallel flag or add
+          # `parallel = True` in the coverage config file.
+          # If using pytest-cov, you can also add the `--cov-append` flag
+          # directly or through PYTEST_ADD_OPTS.
         run: |
-          set -e
-          pytest --cov=./ --cov-report=xml
-          : .coverage confuses coveralls, remove it
-          rm -f .coverage
-      - name: upload coverage data to coveralls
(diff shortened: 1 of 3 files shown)


getzze:
@DerouineauNicolas can you also review this PR please? 
@Diaoul didn't add you as maintainer yet so I cannot assign you automatically :(

DerouineauNicolas:
Looks good to me ! Good job !

getzze:
Thanks!
For the tests, I would like remove the `cassettes` and `data` subfolders before running the tests, run one test to generate the cassettes and then run the tests for the matrix of OS and python versions with the cached cassettes (so the websites are not called 15 times).
Any idea how to do that?

DerouineauNicolas:
I'm not sure this is a good idea to call the external API in the CI. I would leaves the cassettes to catch any requests, but maybe I didn't understand the problematic correctly.

getzze:
My idea was to run automatic test every month, without empty cassettes, so we make sure that the API are still working.
After a PR or push maybe we don't need to call the API and we could used cached cassettes.
But anyway I want the tests for the different OS and python-version to use cached cassettes also.

So maybe writing another test to run every month on a single OS-python-version and with empty cassettes would be OK.

getzze:
For python versions <= 3.9, it seems that the cassettes with a gzipped content cannot be decompressed.
Any idea how to force a non-compression, so we can re-run the faulty cassettes?

getzze:
for python-version 3.10, the tests pass after applying #1127 

for python-version 3.12, we need to solve #1126 first.

DerouineauNicolas:
From my perspective, the simpliest way to disable gzip compression would be to regenerate the cassetes with the option decode_compressed_response=True (https://vcrpy.readthedocs.io/en/latest/advanced.html#decode-compressed-response)

getzze:
Thanks I'll try that

getzze:
all the tests are passing except for one in version py312 which is affected by #1126 .

PR #1120, opened 2024-06-12
Added setuptools as an install dependency

hpsbranco:
enzyme and, by extension, the metadata refiner, relies on the (deprecated) pkg_resources module from setuptools.

This can break the refiner in newly created venvs, since pip doesn't requipre setuptools.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,6 +60,9 @@ target/
 # Pycharm
 .idea
 
+# Visual Studio Code
+.vscode
+
 # Subliminal
 tests/data/mkv/
 tests/data/rar/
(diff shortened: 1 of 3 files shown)


ptrcnull:
this commit doesn't seem to match it's message - instead of adding setuptools, you're commenting it out

ptrcnull:
also, it was a dependency, but it seems it was removed in bdcbffed92329dbe83d4ecbc96383e168555460e (@getzze?)

getzze:
The `develop` branch doesn't need and doesn't depend on `pkg_resources` anymore.

It needs to be fixed in `enzyme`. You should write a PR [there](https://github.com/Diaoul/enzyme/).

ptrcnull:
oh, i missed that enzyme was a separate package

https://github.com/Diaoul/enzyme/issues/33

getzze:
Thanks for opening an issue there.

This PR can be closed.

getzze:
I just realized that the issue was opened more than 6 month ago :sweat_smile:

Issue #1123, opened 2024-06-12
Change CLI default encoding from None to 'utf-8'


getzze:
From https://github.com/Diaoul/subliminal/issues/1112#issuecomment-2156015401

`None` means preserve the original encoding, which means that the video player will not be able to decode it if it's not 'utf-8'.

PR #1125, opened 2024-06-12
Change default subtitle encoding to utf8

getzze:
Before it was the original encoding.

To get the previous behaviour use the flag `--original-encoding` from the cli.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -5,6 +5,7 @@ Changelog
 ^^^^^
 **release date:** *not released*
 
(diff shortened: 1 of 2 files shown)


getzze:
closes https://github.com/Diaoul/subliminal/issues/1123

getzze:
What do you think @ptrcnull ?

ptrcnull (approved):

ptrcnull:
looks good, thanks ^^

PR #1127, opened 2024-06-13
Requires importlib_metadata>=4.6 for tests

getzze:
Only for python_version<3.10

Otherwise one test is failing for these python versions


getzze:
Eventually we could use a `compat.py` file like [babelfish](https://github.com/Diaoul/babelfish/blob/main/babelfish/compat.py)

ptrcnull:
IMHO we're not gonna have enough cases like this to make it a separate file, a few inline `if`s should be alright

PR #1128, opened 2024-06-14
Added setuptools as an install dependency

hpsbranco:
enzyme and, by extension, the metadata refiner, relies on the (deprecated) pkg_resources module from setuptools.

This can break the refiner in newly created venvs, since pip doesn't require setuptools.

It's just a workaround until enzyme is fixed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,6 +60,9 @@ target/
 # Pycharm
 .idea
 
+# Visual Studio Code
+.vscode
+
 # Subliminal
 tests/data/mkv/
 tests/data/rar/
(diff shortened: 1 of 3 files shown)


hpsbranco:
Recreated the PR with the correct pyproject.toml.

I'm working a pull request to fix enzyme.

getzze:
Thanks but I think it's better to wait for the fix in `enzyme`. It only affects python version 3.12 as far as I can tell and a quick fix is to manually install `setuptools`.

hpsbranco:
No problem.

PR #1131, opened 2024-06-22
Bump version to 2.2.0

getzze:
Next steps:
- merge to `master` (I would prefer to use `main` but only @Diaoul can change that, so let's use `master` for this release).
- tag the new version
- make a release on github and pypi (any idea how?)

There is only one breaking change that is the default encoding is now 'utf-8' instead of the original encoding, so I think it's fine not to bump to version 3.0


Diaoul:
> merge to `master` (I would prefer to use `main` but only @Diaoul can change that, so let's use `master` for this release).


Let me know where and when I should do this :+1:

getzze:
> > merge to `master` (I would prefer to use `main` but only @Diaoul can change that, so let's use `master` for this release).
> 
> Let me know where and when I should do this 👍

I just pushed a `main` branch to this repo, you have to make it the default branch in Settings - General - Default Branch. Thanks!

getzze:
I just saw you updated `enzyme`, I'll update the version for the dependency then.

Diaoul:
Yes, please run some tests in case I broke something :laughing:

Diaoul:
@getzze I did something which seems more the way to go
* Changed default branch to develop
* Renamed develop to main

:point_right: this way the PRs are all switched against `main`

getzze:
Ah cool, that's even better!

Tests are passing with the last `enzyme` version 🎉

getzze:
I saw you added a `publish` github action to `enzyme`, if we add it to `subliminal` it will automatically publish the new tagged version to Pypi?

subliminal/cli.py
```diff
@@ -181,7 +181,7 @@ def plural(quantity: int, name: str, *, bold: bool = True, **kwargs: Any) -> str
 )
 @click.option(
     '-c',
-    '--config',
+    '--config-file',
```
ptrcnull, line 184: any reason for the rename? imho `--config` fits better

HISTORY.rst
ptrcnull, line 1: would be neat to also include all dependency changes for downstream packagers

getzze: good point, done!

subliminal/cli.py
getzze: I think it's more accurate because it expects a file path. Why do you prefer `--config` ?

ptrcnull: it feels somewhat more consistent with most other software that i've used, but i guess it's just a personal preference

Dockerfile
getzze, line 1: @DerouineauNicolas I bumped the Docker python version to 3.12, can you check that it's still working please?

ptrcnull:
one little nitpick that i didn't catch earlier: do we have to print `Not using any configuration file.` every single time? wouldn't it be better as a debug print?

subliminal/cli.py
getzze: At the time, I checked other softwares and I had the feeling that `--config-file` was more used :laughing:
Using `--config` makes it more general, so we don't have to change the name if at some point we want to allow `KEY=VALUE` for instance. `ruff` is like that.
`mypy` uses `--config-file`.

getzze: I'll revert then :+1:

getzze:
Yes, I don't like it either.
The problem is that when the config file is parsed, the logger is not set yet therefore the message is "printed".
Another problem I saw is that the two options `--debug` and `--verbose` are redundant.

ptrcnull:
i'd say redundant options are alright, synonymous flags just make it easier for users to get it first try :P

Dockerfile
ptrcnull: i can also confirm that the Docker container works for me

Diaoul:
> it will automatically publish the new tagged version to Pypi?

Yes, not tagged though, released and published on GitHub
1. That usually start with a tag
2. Then you go do make a new release based on that tag (draft or otherwise), you can decide if beta or not
3. Only when the release is "published" (i.e. not draft anymore) the workflow will trigger and publish on PyPI :+1: 

Automation all the way :robot:

getzze:
I added the `publish.yaml` file and a `.readthedocs.yaml` file also. I will draft a release!

ptrcnull:
good job!! thanks everyone, really happy to see a new release ^^

getzze:
I managed to make the release! Thank you all!
I changed the github action to automatically make a Github release and a Pypi release when a new tag is added.
The documentation in readthedocs was not updated though. If found this [gha](https://github.com/peaceiris/actions-gh-pages), but I am not sure it would work.

PR #1137, opened 2024-06-25
Relax the version bounds on platformdirs to `>= 3`

### LINKED ISSUE #1136: FYI: Subliminal 2.2.0 can not be used in Fedora and derivatives.
For compatibility reasons, Fedora uses `virtualenv 20.21.1`, which requires `platformdirs <4,>=2.4`, so Fedora uses `platformdirs 3.11.0`. Therefore `Subliminal 2.2.0` can not be installed/used in Fedora (and many other Linux distributions) because the upgraded requirement to `platformdirs >= 4.2`. Unfortunately Fedora and derivatives use (and will keep using) `platformdirs 3.11.0` to keep compatibility with legacy versions of Python.

Details at [https://src.fedoraproject.org/rpms/python-virtualenv](https://src.fedoraproject.org/rpms/python-virtualenv). See the comments in `python-virtualenv.spec` file.

ptrcnull:
fixes #1136

as a side note, should we update `HISTORY.rst` with every single change, under a "not released yet" header? it might make things easier on release, as we won't have to go through all the commits to check what was changed


Diaoul:
> as a side note, should we update `HISTORY.rst` with every single change, under a "not released yet" header? it might make things easier on release, as we won't have to go through all the commits to check what was changed

I would suggest to remove this file and use a generator to craft GitHub releases instead. I don't have recommendations as I don't know those tools but I'm pretty sure some popular projects use them (maybe FastAPI?)

getzze:
I saw some project using [`towncrier`](https://towncrier.readthedocs.io/en/stable/) but I have never used it myself.

In the meantime, I think it's a good idea to write `HISTORY.rst` for each commit, good point.

ptrcnull:
> I would suggest to remove this file and use a generator to craft GitHub releases instead. I don't have recommendations as I don't know those tools but I'm pretty sure some popular projects use them (maybe FastAPI?)

IMHO it's way more useful to have a changelog file, since it's not tied to GitHub and you can just look it up when you have the repo checked out locally

ptrcnull:
> I think it's a good idea to write `HISTORY.rst` for each commit

i'd say each merge request? ideally so that the changelog makes sense for users who don't exactly have to know *how* Subliminal actually works

PR #1139, opened 2024-06-26
Add example subliminal.toml to docs and various docs fixes

getzze:
The main fix is that the members of the modules and classes were not showing in the docs.


getzze:
@Diaoul can you check why Readthedocs is not showing the latest version?

Diaoul:
Sure will do!

Diaoul:
[Fixed](https://subliminal.readthedocs.io/en/2.2.0/) (wrong branch configured)

Issue #1138, opened 2024-06-25
Automate releases


getzze:
Several things could be done to ease the process of making a new release:

- [ ] protect the `main` branch and review the github actions to ensure the linter and tests pass before a release.
- [ ] use `pre-commit.ci` to automatically fix linter and formatting problems https://github.com/Diaoul/subliminal/issues/1075#issuecomment-2164144583
- [x] show a coverage report on coverage failure
- [x] automatic building of changelog (using `towncrier` like [attrs](https://github.com/python-attrs/attrs) ?)

wujekbogdan:
Have you considered [Python Semantic Release](https://github.com/python-semantic-release/python-semantic-release?tab=readme-ov-file)? It's inspired by the great [semantic-release](https://github.com/semantic-release/semantic-release) tool for JS.

getzze:
I didn't know it, thanks. I looked at the documentation and I found it quite confusing. From what I understood it uses commit messages to decide of the next version? But does it automate the release on Github?

At the moment I'm thinking of a solution based on setuptools_scm and python scripts to do that.

wujekbogdan:
You're getting it right. The idea is that the tool determines the version number based on commit messages. It's often implemented in such a way that you squash-merge a PR and use its commit message as a source of truth.

Then you use that information as input for a GitHub Action that updates the GitHub tag and performs the actual release. See: https://python-semantic-release.readthedocs.io/en/latest/automatic-releases/github-actions.html

PR #1142, opened 2024-07-03
Use towncrier to avoid conflicts in HISTORY.rst

getzze:
See #1138

Documentation [here](https://towncrier.readthedocs.io/en/stable/). It is used by many many projects (attrs, pytest, numpy, ...) so it seems a good choice.

Like that, with each commit or PR, we add a news fragment (a file) in the `changelog.d` folder.
Before making a new release, we call `towncrier -- version 2.3.0` and the files are parsed and added nicely to the `HISTORY.rst` file. They are also deleted from the `changelog.d` folder.
You can add a news fragment from the command line with: `towncrier create --content 'use towncrier' 1138.misc.rst`

We have to decide the type of fragments to add. I put:
 - `breaking`
 - `change`
 - `provider`
 - `refiner`
 - `deprecation`
 - `doc`
 - `misc`

With `doc` and `misc` showing only the list of PR numbers in the changelog, not the text.

Notice that it will change the formatting of the `HISTORY.rst` file, now the title for a new release with the date will be on the same line.

I will add the information about how to make a new release in another PR (first handle the conflicts in `HISTORY.rst`).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -31,3 +31,12 @@ repos:
       - id: doc8
         description: This hook runs doc8 for linting docs
         require_serial: true
+
+  - repo: local
+    hooks:
+    - id: changelog-fragment-filenames
+      name: changelog fragment
+      language: fail
+      entry: changelog fragment files must be named *.(breaking|change|provider|refiner|deprecation|doc|misc).rst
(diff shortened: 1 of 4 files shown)


Diaoul:
Cool tool, did not know this one. Out of curiosity, what are alternatives? Why this one and not another?

getzze:
I didn't check other tools as this one looked great...

I found this [talk](https://hynek.me/talks/python-foss/) by the `attrs` author about packaging and I started checking how they do it in `attrs`, `pytest` and `numpy`.

PR #1144, opened 2024-07-04
Update README with newer infos

getzze:
Update badges and installation method.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -22,6 +22,10 @@ Subtitles, faster than your thoughts.
     :alt: Join the chat at https://gitter.im/Diaoul/subliminal
     :target: https://gitter.im/Diaoul/subliminal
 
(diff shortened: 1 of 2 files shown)


ptrcnull (changes requested):
two little things:
- discord badge: `[Discord](https://discord.gg/kXW6sWte9N)` ( @Diaoul, I generated a new invite code because yours was limited to 1 year; probably not a problem, but we can make an unlimited one so why not )
- maybe we should make the towncrier snippets capitalized?

getzze:
FIY, the url of the discord badge gives an Invalid badge.

ptrcnull:
ah, edited with the correct URL

getzze:
> * maybe we should make the towncrier snippets capitalized?

Maybe this can be enforced at build time.

PR #1150, opened 2024-07-06
OpenSubtitlesCom: avoid duplicate subtitles

### LINKED ISSUE #1146: Too many results from opensubtitlescom provider
First of all, thank you for taking up this repo again.
While I was migrating to the newest version (2.2.1), I noticed that the opensubtitlescom provider returns too many results.
image

When enabling the debug logs, it seems that multiple requests are made?
image

Any idea why this is done?
Because there are actually only 3 records available for the provider.
image

Shouldn't it be limited to 1 call?
If not, then I would propose to filter out the duplicate subtitles (you can that by checking the id) before returning the result.

getzze:
fixes #1146

(the change in `providers/__init__.py` is to pass `ruff check` with ruff 0.5.1

### DIFF
--- a/subliminal/providers/opensubtitlescom.py
+++ b/subliminal/providers/opensubtitlescom.py
@@ -675,7 +675,7 @@ class OpenSubtitlesComProvider(Provider):
         # fill the search criteria
         criteria = self._make_query(**kwargs)
 
-        subtitles = []
+        subtitles: list[OpenSubtitlesComSubtitle] = []
 
         for criterion in criteria:
             # add the language and query the server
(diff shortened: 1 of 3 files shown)


subliminal/providers/opensubtitlescom.py
```diff
@@ -695,8 +695,12 @@ def query(
                     imdb_match=imdb_match,
                     tmdb_match=tmdb_match,
                 )
-                logger.debug('Found subtitle %r', subtitle)
-                subtitles.append(subtitle)
+
+                # Some criteria are redundant, so skip duplicates
+                unique_ids = [s.id for s in subtitles]
```
Diaoul, line 700: Use a set for O(1) lookups

```suggestion
                unique_ids = {s.id for s in subtitles}
```

Diaoul (approved):

PR #1154, opened 2024-07-15
Use knowit to parse video metadata (frame_rate, duration, subtitles...)

getzze:
Using [knowit](https://github.com/ratoaq2/knowit) has two advantages:
- it can read the frame rate of the video (mkv or other) to be used to match subtitles and to convert frame-based subtitles to srt.
- it gives more information about subtitle tracks than enzyme and it should find subtitles in .mov files.

In terms of dependency, it needs or `ffmpeg` or `mediainfo` installed, otherwise it fallbacks on using `enzyme`.
It's maintained by @ratoaq2 who used to maintain subliminal, so it fits very well :)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,12 +20,16 @@ if [ "$BUILD_WITH_UNRAR" = true ]; then \
     rm -rf /tmp/unrar /tmp/unrar.tar.gz; \
 fi
 
+# install libmediainfo for metadata refiner
+RUN apk add --no-cache libmediainfo
+
 RUN mkdir -p /usr/src/app /usr/src/cache
 
 WORKDIR /usr/src/app
 VOLUME /usr/src/cache
 
 COPY . /usr/src/app
+RUN python -m pip install -U pip
 RUN python -m pip install .
 
 
--- a/README.rst
+++ b/README.rst
@@ -91,6 +91,8 @@ clone your fork locally and install a development version::
     $ cd subliminal
     $ pip install --user -e '.[dev,test,docs]'
 
+To extract information about the video files, `subliminal` uses `knowit <https://github.com/ratoaq2/knowit`.
+For better results, make sure one of its provider is installed, for instance `MediaInfo <https://mediaarea.net/en/MediaInfo>`.
 
 Integrations
 ------------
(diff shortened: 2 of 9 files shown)


getzze:
test are failing on macos and windows because `mediainfo` is not installed.

failing on python 3.12 because https://github.com/ratoaq2/knowit/pull/176 is not merged yet.

Diaoul:
As long as there is no hard dependency to install ffmpeg or all those stuff, I'm fine with it!

💡 Might want to add those in the docker image as well

getzze:
`knowit` depends on `pymediainfo`, but it's a Python package. The `libmediainfo` library is bundled in the wheel package on Windows and MacOS. On Linux you need to install it manually. I should document it.

There is currently a bug in `knowit` so the bundled library is not automatically detected. It needs to be solved before it can be used in `subliminal`.

> Might want to add those in the docker image as well

Good point, thanks

Diaoul:
I think this would make subliminal a non pure-python package now? That's OK, packaging is much easier nowadays with wheels and all.

But there might be some people running it on embedded platforms 😅

getzze:
If you don't have mediainfo, ffmpeg or mkvmerge external libraries installed, `knowit` falls back to `enzyme`. So external dependencies are completely optional.

getzze:
Tests passing finally!

I opened a PR at [pymediainfo](https://github.com/sbraz/pymediainfo/pull/138) to include the library on Linux and MacOS-arm64 (currently it's only bundled for Windows and MacOS-x86_64). So people would not have to install any external library to improve the parsing.

I think I should add a small paragraph to the docs (the one from the news fragment maybe). But I don't know where it would be best, any idea?

nunotexbsd:
@getzze 

For what I understand, kwowit will be a test depend?
If so, I will need to port it to FreeBSD then.

getzze:
`knowit` will be a mandatory dependency to `subliminal`, it's a python package.
It has 4 providers, 2 being completely external (non-python) soft dependencies, `ffmpeg` and `mkvmerge`; one provider is
`pymediainfo`, which is a python package, that is a wrapper for the external library `libmediainfo`. However `pymediainfo` wheel on Pypi has the library bundled (so you don't have to install it yourself), but only for Windows and old Mac. It's not possible to bundle it for FreeBSD. But it's still possible to install `libmediainfo` from the package manager. The `python-pymediainfo` package on Arch Linux has a dependency to `libmediainfo` [for instance](https://archlinux.org/packages/extra/any/python-pymediainfo/).

The last provider is the python package `enzyme` that was a dependency of `subliminal`. So even if no external tool is installed, it will still work as before (no improvement though).

Issue #995, opened 2020-03-29
suggestion: use ctime to determine file age for filtering


ittayd:
If someone moves files from one folder to the folder subliminal scans, their modification time doesn't change. Therefore the -a flag might cause subliminal to ignore them. It would be good to instead or in addition use ctime.

getzze:
Duplicate of #860

PR #1162, opened 2024-07-29
Allow to use the latest of creation and modification date for age

### LINKED ISSUE #860: Age filtering when scanning files should use creation date
Currently they are filtered by modified date [here](https://github.com/Diaoul/subliminal/blob/develop/subliminal/core.py#L546), but doesn't creation date make more sense? 

If you put/copy new files into the scanned folder they should be treated as new. The modified date will never change unless you re-encode, and that would probably also change the hash and break subtitle detection.

The creation date is platform dependent but I found [this](https://stackoverflow.com/a/39501288) stackoverflow answer with cross-platform code.

edit: Probably better to use creation and modified dates and pick the latest?
So just `max(os.path.getmtime(path), os.path.getctime(path))`

getzze:
fixes #860, #995

### DIFF
--- a/changelog.d/860.change.rst
+++ b/changelog.d/860.change.rst
@@ -1,0 +1,1 @@
+Allow to use the latest of creation and modification date to compute the age of the file.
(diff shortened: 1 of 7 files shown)


getzze:
I'm wondering if using the latest of creation date and modification date should be the default behavior (instead of using only modification date).

Issue #692, opened 2016-09-27
Add support for Air By Date releases


duramato:
Currently when searching for subs for a release with air by date numbering , subliminal just errors
Like so:
```
subliminal --debug download -l en Panorama.2016.06.29.Iraq.The.Final.Judgement.720p.WEB-DL.h264-OmarMateen.mkv
Collecting videos  [####################################]  100%  Panorama.2016.06.29.Iraq.The.Final.Judgement.720p.WEB-DL.h264-OmarMateen.mkvDEBUG:subliminal.cli:Collecting path Panorama.2016.06.29.Iraq.The.Final.Judgement.720p.WEB-DL.h264-OmarMateen.mkv
INFO:subliminal.core:Scanning video 'Panorama.2016.06.29.Iraq.The.Final.Judgement.720p.WEB-DL.h264-OmarMateen.mkv' in ''
ERROR:subliminal.cli:Unexpected error while collecting path Panorama.2016.06.29.Iraq.The.Final.Judgement.720p.WEB-DL.h264-OmarMateen.mkv
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/subliminal/cli.py", line 352, in download
    video = scan_video(p)
  File "/usr/local/lib/python2.7/dist-packages/subliminal/core.py", line 380, in scan_video
    video = Video.fromguess(path, guessit(path))
  File "/usr/local/lib/python2.7/dist-packages/subliminal/video.py", line 93, in fromguess
    return Episode.fromguess(name, guess)
  File "/usr/local/lib/python2.7/dist-packages/subliminal/video.py", line 166, in fromguess
    raise ValueError('Insufficient data to process the guess')
ValueError: Insufficient data to process the guess
Collecting videos  [####################################]  100%
0 video collected / 0 video ignored / 1 error
```

It shoud also search for them .

ratoaq2:
This is not that simple since the providers requires episode information. Even tvdb refiner can only search episode information with the episode number.

PR #702, opened 2016-10-12
Allow episode search without episode number

evuez:
This fixes #692.

Not finding the episode number shouldn't raise an error, most of the time
it actually works without one anyway.

Issue #678, opened 2016-07-27
Season match inconsistency


ratoaq2:
For the latest version: 2.0.3

There's a season match inconsistency when release names have no season (I'll provide real examples later today). 

Given a video file as bellow:
`Show.Name.E05.720p.x264.mkv`

The created video object will have season 1, since guessit will not return season, but this [code](https://github.com/Diaoul/subliminal/blob/master/subliminal/video.py#L168) will default it to 1.

Then when the subtitles (e.g.: `Show.Name.E05.720p.x264,srt`) are found (by any provider), some/most of them will use guessit to retrieve all the release information and this [code](https://github.com/Diaoul/subliminal/blob/master/subliminal/subtitle.py#L207) will prevent a season match, since `video.season == 1` but `'season' is not in guess`

Either the season shouldn't be defaulted to 1 in the video or the compute score should follow the same logic.

Diaoul:
I agree on that. The code you linked to was introduced to handle mini-series that don't have a season number in the filename. Something more clever is needed here.

ratoaq2:
Just to illustrate and give some real examples, there are series that doesn't follow season/episode numbering on their release names. Not all providers have subtitles for those series.
Here are some examples of episodes without season and/or episode numbers:

- Air by date (guessit: `date`): documentaries episodes or talk shows
`Catalyst.2015.11.10.Einstein.360p.LDTV.WEBRIP.[MPup]`
`Eastenders.2016.05.27.HDTV.x264-DEADPOOL`

- Absolute episode numbering (guessit: `episode`): animes
`Kiznaiver.E06.SUBFRENCH.720p.HDTV.x264-D4KiD`

- Absolute episode using `part` (guessit: `part`):
`The.Night.Of.Part.8.INTERNAL.720p.HDTV.x264`

PR #1165, opened 2024-08-02
Remove requirement for episode number when guessing

ptrcnull:
with some types of episodes, such as extras, the episode number will be missing, but the subtitle download succeeds with the check removed

### DIFF
--- a/changelog.d/+4eebcb09.change.rst
+++ b/changelog.d/+4eebcb09.change.rst
@@ -1,0 +1,1 @@
+Remove requirement for episode number when guessing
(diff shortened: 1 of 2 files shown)


ptrcnull:
ah, actually, sometimes it just pulls the first episode of the season...

getzze:
Fixes #692

And it would be great if this PR could also tackle #678

They don't look easy to fix though...

getzze:
Are you still working on this?
I just realized it's a duplicate of: https://github.com/Diaoul/subliminal/pull/702

ptrcnull:
to be fair, the other MR is flawed in the same way as mine is

ptrcnull:
but to answer your question, i forgot about it, but *at some point* i'd like to either fix it myself, or see it fixed :)

PR #826, opened 2017-10-30
New spanish providers: www.tusubtitulo.com, www.subtitulamos.tv

porzino:
Added tusubtitulo.com and subtitulamos.tv providers.

PR #1170, opened 2024-09-22
Add Subtitulamos provider integration

Nyaran:
Added Subtitulamos provider

This PR is partially based and replaces #826 (skipping the provider "tusubtitulo" as is not working anymore)

~~Also I added a way to run the cli from source code ([subliminal/cli.py#L727-L728](https://github.com/Nyaran/subliminal/blob/6699c62f221d2e6392f86fed008265a01d5bc316/subliminal/cli.py#L727-L728)) , as is helpful to debug new additions (if you don't like it, tell me and I'll remove it)~~

Pending:

- [x] Add tests
- [x] Changelog entry
- [x] CI passes (except docs)

### DIFF
--- a/subliminal/cli.py
+++ b/subliminal/cli.py
@@ -39,8 +39,7 @@ from subliminal import (
     scan_videos,
 )
 from subliminal.core import ARCHIVE_EXTENSIONS, scan_name, search_external_subtitles
-from subliminal.extensions import default_providers, default_refiners
-from subliminal.score import match_hearing_impaired
+from subliminal.extensions import get_default_providers, get_default_refiners
 from subliminal.utils import merge_extend_and_ignore_unions
 
 if TYPE_CHECKING:
@@ -142,6 +141,12 @@ def configure(ctx: click.Context, param: click.Parameter | None, filename: str |
 
     # make download options
     download_dict = toml_dict.setdefault('download', {})
+    # handle language types
+    for lt in ('hearing_impaired', 'foreign_only'):
+        # if an option was defined in the config file, make it a tuple, the expected type
+        if lt in download_dict and (isinstance(download_dict[lt], bool) or download_dict[lt] is None):
+            download_dict[lt] = (download_dict[lt],)
+
     # remove the provider and refiner lists to select, extend and ignore
     provider_lists = {
         'select': download_dict.pop('provider', []),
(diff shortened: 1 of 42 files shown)


subliminal/cli.py
```diff
@@ -722,3 +722,7 @@ def download(
 
     if verbose == 0:
         click.echo(f"Downloaded {plural(total_subtitles, 'subtitle')}")
+
+
+if __name__ == '__main__':
+    subliminal()
```
ptrcnull, line 728: this should be in a separate commit, if anything

ptrcnull: also, if this was just a temporary launcher thingy for development, you can achieve the same with `python3 -m subliminal`

Nyaran: I tried that, and I get " No module named subliminal.__main__; 'subliminal' is a package and cannot be directly executed" (I'm running from the virtual env, after installing dependencies)

Anyway, no problem, before I finish this PR I will reverse the change

getzze:
Thanks for updating this provider!

Don't worry too much about mypy (or if you need help let me know). It's better if coverage is high ;)

subliminal/cli.py
getzze: You can install subliminal in editable mode with `pip install '.[dev]' -e`, even in a virtual env. Like that you can install subliminal and run it's CLI from the terminal, and at the same time modify the files and see the modification in the installed package.

subliminal/providers/subtitulamos.py
```diff
@@ -0,0 +1,310 @@
...
+            language = Language.fromsubtitulamos(lang_name_element.get_text().strip())
+
+            hearing_impaired = False
+
+            # modify spanish latino subtitle language to only spanish and set hearing_impaired = True
+            # because if exists spanish and spanish latino subtitle for the same episode, the score will be
+            # higher with spanish subtitle. Spanish subtitle takes priority.
+            if language == Language('spa', 'MX'):
+                language = Language('spa')
+                hearing_impaired = True
```
getzze, line 238: this is a hack that will stop working with changes to come to how `hearing_impaired` and `foreign_only`/`forced` subtitles are sorted. See #1175 for the first step.

Why do you need to transform a american-spanish subtitle to a european-spanish subtitle? It can be specified by having both languages (or only one) in the command line, no?

Nyaran: I'm not really sure, I thought the same when I saw that on the original proposal, but I decided to keep it in case there was something I missed. Let me perform some test and make the commit removing the block.

Nyaran: Done, and after reviewing it I improved this to support all countries with Spanish language (all of them under the name "Español (Latinoamérica)" on the provider.

subliminal/providers/subtitulamos.py
```diff
@@ -0,0 +1,342 @@
...
+                or (lang_name_element := sub.find_previous('div', class_='language-name')) is None
+                or (version_container := sub.find_previous('div', class_='version-container')) is None
+                or not isinstance(version_container, Tag)
+                or (release_group_element := version_container.select('.version-container .text.spaced')) is None
+            ):
+                continue
+
+            language = Language.fromsubtitulamos(lang_name_element.get_text().strip())
+
+            hearing_impaired = False
```
getzze, line 259: Do we need to force `hearing_impaired = False`, meaning we are sure they ARE NOT hearing impaired? The default, `None`, means we don't know.

subliminal/providers/subtitulamos.py
getzze: Thanks!

subliminal/providers/subtitulamos.py
Nyaran: I think subtitulamos doesn't have hearing impaired subtitles, but I'm not sure, so... line removed

getzze: ok, thanks

getzze:
Great job!

PR #1175, opened 2024-10-07
Preference or avoidance per language type (hearing impaired, foreign only)

### LINKED ISSUE #535: Detect hearing impaired subtitles when the provider doesn't give the information
Hey.

Usually when I run: 

```
subliminal download -l en movie.mp4
```

It will come with hearing impaired subtitles. I find this very frustrating. Can't there be a introduction of `--avoid-hearing-imparied`?

The goal of that flag should be that it will try to find non-hearing-impaired subtitles first, and if they can't be found it's OK to pick a HI subtitle - but only as a last resort.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28368588-detect-hearing-impaired-subtitles-when-the-provider-doesn-t-give-the-information?utm_campaign=plugin&utm_content=tracker%2F321252&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F321252&utm_medium=issues&utm_source=github).
</bountysource-plugin>

getzze:
closes #535
partly solves #925

Currently, we can only choose to prefer hearing impaired subtitles (among subtitles with the same highest score) with the "-hi" option. This PR adds the following:
 - Add cli option "-HI" to prefer non-hearing-impaired subtitles, negation of existing "-hi".
 - Add cli options "-fo" and "-FO" to prefer (or disfavor, respectively) foreign-only/forced subtitles.

 "-hi" takes precedence over "-fo" and "-FO" (it cannot be used with "-HI").
 "-fo" takes precedence over "-HI" (it cannot be used with "-FO").
 Using "-HI" and "-FO" together is redundant and equivalent of using just one of the options.

However, not all providers are reporting foreign_only tags for subtitles (added opensubtitles.com support in this PR).

Other:
 - Rename "forced" to "foreign_only" from #1148

### DIFF
--- a/changelog.d/1175.change.rst
+++ b/changelog.d/1175.change.rst
@@ -1,0 +1,1 @@
+Add cli option to prefer or avoid hearing impaired or foreign only subtitles.
--- a/changelog.d/1175.change.rst
+++ b/changelog.d/1175.change.rst
@@ -1,1 +1,1 @@
-Add cli option to prefer or avoid hearing impaired or foreign only subtitles.
+Add cli option to prefer or disfavor hearing impaired (-hi/-HI) or foreign only (-fo/-FO) subtitles.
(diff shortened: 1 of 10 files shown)


getzze:
Also I didn't modify the scores, I will do it in another PR when adding an fps match.

PR #1176, opened 2024-10-07
Add python3.13, drop python3.8

getzze:


getzze:
add the change proposed in https://github.com/Diaoul/subliminal/pull/1150 comment and that was forgotten.

Issue #1177, opened 2024-10-10
Example: How to create and register a subliminal provider extension.


etherealite:
If you are like me, you want to make your own provider but you don't want to fork or wrap subliminal.

For anyone wondering how to make their own provider 'extension', you may have noticed that documentation for this is a placeholder:
[https://subliminal.readthedocs.io/en/latest/api/extensions.html](https://subliminal.readthedocs.io/en/latest/api/extensions.html)
```markdown
**Extensions
Extension managers for the providers, refiners and language converters.
```

But don't fret, after many trials and tribulations I have figured out how to perform the magical incantations necessary to make an extension. Let my own suffering be your salvation and read on...

All you have to do is make an entry in your pyproject.toml like so:
```toml
[project.entry-points."subliminal.providers"]
dummy = "ihatestevedore.dummyprovider:DummyProvider"
```

In this example, your package name is `ihatestevedore` and it has a module named `dummyprovider` containing your provider class definition. The entry name that you assign (left side of the '=') is arbitrary and can be whatever you want. You can then create your package layout like so:
```
/path/to/extension/ihatestevedore
└───ihatestevedore
    │   dummyprovider.py
```

Where you implement the `DummyProvider` class in the dummyprovider module:
```python
from subliminal import Provider

class DummyProvider(Provider):
    pass
```

Here is what a complete project.toml might look like:

```toml
[project]
name = "ihatestevedore"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
requires-python = ">=3.12"
dependencies = [
    "subliminal>=2.2.1",
]

[project.entry-points."subliminal.providers"]
dummy = "ihatestevedore.dummyprovider:DummyProvider"

[build-system]
requires = ["hatchling"]
build-backend = "hatchling.build"
```

After installing your package, there should be an `entry_points.txt` within its home in the site-packages directory:
```
$ ls python/Lib/site-packages/ihatestevedore-0.1.0.dist-info
direct_url.json
entry_points.txt
INSTALLER
METADATA
RECORD
REQUESTED
WHEEL
```

It should look like:
```toml
[subliminal.providers]
dummy = ihatestevedore.dummyprovider:DummyProvider
````

Now your extension should be available for discovery by subliminal and you can follow the rest of the docs and examples concerning making your own provider.

More reading on how this works:
* making plugins with stevedore https://docs.openstack.org/stevedore/latest/user/tutorial/creating_plugins.html
[...]

PR #1180, opened 2024-10-17
Cli providers params

### LINKED ISSUE #1179: Pass arguments to providers through the CLI
Right now we can pass arguments to the providers only through the config file.
Because of the extension system, we cannot set the arguments of each and every providers as hard-coded options for the CLI. However, `click` allows to add dynamic options!

The scheme for the provider (and refiner) arguments I thought about is:
`--provider.opensubtitles.username=user`

The variable names are `_provider__opensubtitles__username`, note the double underscores to be able to split the variables if an underscore is present in an argument name or provider name, for instance `_provider__addic7ed__allow_searches`.

We could also add support for passing options to the CLI with environment variables (easy to do with `click`).
The scheme would be:
`SUBLIMINAL_PROVIDER_OPENSUBTITLES_USERNAME` (with single underscores).

getzze:
closes #1179 

This is a breaking change, for command line usage.

### DIFF
--- a/changelog.d/1179.breaking.rst
+++ b/changelog.d/1179.breaking.rst
@@ -1,0 +1,4 @@
+Deprecate the `--addic7ed USERNAME PASSWORD`, `--opensubtitles` and `--opensubtitlescom` CLI options
+in favor of `--provider.addic7ed.username USERNAME`, `--provider.addic7ed.password PASSWORD`, etc...
+Add a generic way of passing arguments to the providers using CLI options.
+Use environment variables to pass options to the CLI.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -88,7 +88,7 @@ repository = "https://github.com/Diaoul/subliminal"
 documentation = "https://subliminal.readthedocs.org"
 
 [project.scripts]
-subliminal = "subliminal.cli:subliminal"
+subliminal = "subliminal.cli:cli"
 
 [project.entry-points."subliminal.providers"]
 addic7ed = "subliminal.providers.addic7ed:Addic7edProvider"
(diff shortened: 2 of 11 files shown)


etherealite:
I like a lot of what I'm seeing here, getting rid of the click-option-group package is nice. Keeping all of the click usage confined to one module is clean.

 I think I would prefer if this could be made a little bit closer to the approach that the Typer project uses. It combines using docstrings with an annotated function signature. See it in action in the [CLI Options with Help](https://typer.tiangolo.com/tutorial/options/help/) section of their docs.

I would really really really prefer to not have anything syntactic in the doc strings. Using the doc string in the generated docs is great, but don't parse it.

```python
class SomeProvider(Provider):
    def __init__(
            self,
            username: Annotated[str, Option(help="OpenSubtitles username")],
	    password: Annotated[str, Option(help="OpenSubtitles password")],
	)
        """
        Downloads subtitles from OpenSubtitles.
        
        [username] and [password] are required if not set in config.toml
        """
        ...
```

Instead, use the signature annotations for anything where you'd need to parse a docstring and inline the docstring into the generated documentation.


I think it's worth raising that hard coding the parameters that an extension can use is pretty restrictive. Who knows what kind of requirements might be needed by some future extension. Is there not some way to allow more flexibility in this regard? 

Perhaps allowing aping a narrow subset of click's api and exposing that to extensions would work?

getzze:
I like the usage of Annotated type hints, thanks, it's much better than parsing the docstring.

> I think it's worth raising that hard coding the parameters that an extension can use is pretty restrictive. Who knows what kind of requirements might be needed by some future extension. Is there not some way to allow more flexibility in this regard?

It's not the case! It takes all the arguments of the `__init__` method and transform them into `click` options. I hard-coded arguments to SKIP for the refiners but not the providers.

etherealite:
I'll try and incorporate this and the examples you gave in #1177 into my provider soon.

getzze:
I decided not to use `Annotated` type hint and parse the docstring. If needed we can do it in another PR.

PR #1188, opened 2024-11-22
Use hatch and hatch-vcs

### LINKED ISSUE #1076: Automatic versioning
We could use something like `setuptools_scm` but then we cannot install from the tarballs.
An alternative is using `bump-my-version`: (https://github.com/callowayproject/bump-my-version/)

getzze:
closes #1076 

* Version is tracked with `setuptools_scm` via git tags, ~~with a `_version.py` file created when building the package~~ (it's a mess, no `_version.py` file is created).
* The source files were moved to `src/subliminal` to avoid errors while testing.
* Use hatch environments instead of tox for the github actions.
* Update versions of pre-commit hooks.
* Rename the optional dependencies `test` -> `tests`. So to install subliminal with the test dependencies, you need to use `pip install subliminal[tests]`


getzze:
this PR is too big and should be split into smaller PRs.

PR #1197, opened 2024-12-11
Feature: Add __main__.py for using with python -m

glensc:
This allows easily run the program from source tree after installing dependencies:

```
$ pip install -e .
$ python -m subliminal --help
```

### DIFF
--- a/src/subliminal/__main__.py
+++ b/src/subliminal/__main__.py
@@ -1,0 +1,20 @@
+from __future__ import annotations
+
+if len(__package__) == 0:
+    import sys
+
+    print(
+        f"""
+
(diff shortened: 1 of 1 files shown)


getzze:
Thanks for the PR! It looks good to me, just need to fix ruff warnings:

* ruff automatically gets rid of `print` statements, just add this line as the first line of the file.
```
# ruff: noqa: T201
```

* you need to add a docstring to the file.

glensc:
what does it mean, gets rid of print statements, what should one use then?

getzze:
It is to remove `print` statement that have been added for debugging and forgot to remove them.
It can be quite useful :)

If you really need a `print` (like here), you can add an exception to ruff.

Issue #1196, opened 2024-12-10
Missing arm64 docker image


glensc:
```
$ docker run --rm --name subliminal -v subliminal_cache:/usr/src/cache -v `pwd`:/app -w /app -it diaoulael/subliminal download -l en blah.avi
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
exec /usr/local/bin/subliminal: exec format error
```

ptrcnull:
i suppose we could build those? i'm using subliminal on arm64 and it works just fine, though i'm not sure how to wire GH Actions to building cross-platform yet

glensc:
it should be pretty straight forward to add docker image publishing from github workflow.

The existing Dockerfile doesn't seem to be specific to x64:
- https://github.com/Diaoul/subliminal/blob/da2817da2cee503cdc7f4bcc9595d7ecefae4a23/Dockerfile

I can send a PR if there's no obvious reason pointed out why that would be rejected.

glensc:
Current, local workaround, build the image locally:

```sh
docker build -t diaoulael/subliminal https://github.com/Diaoul/subliminal.git#HEAD
```

glensc:
Something not right yet


```diff
diff --git a/Dockerfile b/Dockerfile
index 60f3903..9ff7064 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,6 +30,7 @@ VOLUME /usr/src/cache
 
 COPY . /usr/src/app
 RUN python -m pip install -U pip
+RUN apk add git
 RUN python -m pip install .
 
 
❯ vi ./.dockerignore
❯ g diff
diff --git a/.dockerignore b/.dockerignore
index b77cc43..2547808 100644
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,4 @@
 # SCM
-.git/
 
 # Byte-compiled / optimized / DLL files
 __pycache__/
diff --git a/Dockerfile b/Dockerfile
index 60f3903..9ff7064 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,6 +30,7 @@ VOLUME /usr/src/cache
 
 COPY . /usr/src/app
 RUN python -m pip install -U pip
+RUN apk add git
 RUN python -m pip install .
 
 

```

had to make those local changes for build to succeed.

glensc:
@ptrcnull It's pretty straight forward, and imho well documented

[...]

PR #1248, opened 2025-04-02
Improve dockerfile with best practices

glensc:
_Part of https://github.com/Diaoul/subliminal/issues/1196_

See commit messages for details:
* Dockerfile: Skip copying all sources to docker image
* Dockerfile: Do not include /root/.cache/pip in docker image
* Dockerfile: Do not upgrade pip
* Dockerfile: Remove un-needed absolute path
* Dockerfile: Remove un-needed mkdir
* Dockerfile: Add git for setuptools-scm

PS: I prefer that my PR is not squash-merged, I created my commits atomic:
- https://www.freshconsulting.com/insights/blog/atomic-commits/

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,6 +1,3 @@
-# SCM
-.git/
-
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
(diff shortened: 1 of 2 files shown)


getzze:
Thanks!
`Dockerfile` was not updated when we switched to using `hatch-vcs` for versioning.

PR #1251, opened 2025-04-05
GHA: Create docker build workflow

glensc:
This builds on every pull request,
but publishes only on main branch and git tags

Additionally semver compatible base tags are created for docker images.

The images are published to ghcr.io registry (simpler).

Refs
- https://github.com/Diaoul/subliminal/issues/1196

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -23,6 +23,8 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
+        with:
+          persist-credentials: false
 
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v3
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -31,6 +31,8 @@ jobs:
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
+        with:
+          cache-binary: false
 
       - name: Log in to the Container registry
         uses: docker/login-action@v3
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -1,0 +1,71 @@
+name: Publish Docker image
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - "*.*.*"
+  pull_request:
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+  TEST_IMAGE: subliminal
+
+jobs:
+  docker:
+    runs-on: ubuntu-latest
+    permissions:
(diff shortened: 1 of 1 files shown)


glensc:
I don't know what the CI validation is asking me to do

```
  --> .github/workflows/docker-build.yml:24:9
   |
24 |         - name: Checkout repository
   |  _________-
25 | |         uses: actions/checkout@v4
   | |_________________________________- does not set persist-credentials: false
   |
   = note: audit confidence → Low

error[cache-poisoning]: runtime artifacts potentially vulnerable to a cache poisoning attack
  --> .github/workflows/docker-build.yml:3:1
   |
 3 | / on:
 4 | |   push:
...  |
 8 | |       - "*.*.*"
 9 | |   pull_request:
   | |_______________^ generally used when publishing artifacts generated at runtime
10 |
...
30 |         - name: Set up Docker Buildx
31 |           uses: docker/setup-buildx-action@v3
   |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cache enabled by default here
   |
   = note: audit confidence → Low

54 findings (1 ignored, 51 suppressed): 0 unknown, 0 informational, 0 low, 1 medium, 1 high
```

glensc:
Documentation about that specific error (cache-poisoning):
- https://woodruffw.github.io/zizmor/audits/#cache-poisoning

glensc:
uh, whatever: 2447b8080ba9e2313dc9e8eb0eb85ef66a91d9f3

getzze:
> Documentation about that specific error (cache-poisoning):
> 
>     * https://woodruffw.github.io/zizmor/audits/#cache-poisoning

I think the problem is that the same `docker` job is run after a PR (insecure) and after a push (secure). Therefore, an attacker can poison the cache with a PR and the same cache will be used for a push. Instead of disabling caching, I think it would be better to write a job for PRs (and push) that just check that it builds, [like this (up to "call: check")](https://docs.docker.com/build/ci/github-actions/checks/). Then if this job passes and only for push, the container is built and sent to docker and ghcr.io.

getzze:
I found out that the build can be published on both ghcr.io and dockerhub:
https://docs.docker.com/build/ci/github-actions/push-multi-registries/

But maybe it doesn't make sense to have it in 2 places.

glensc:
If you want to do more cleanup or do the cache thing better, you can do that after this is merged. this PR is ready from my side.

How I see it in the big picture:
1. you merge this
2. make a release
3. you understand how images are published (main, 2.x, 2.2.x, latest)
4. update documentation, include info about docker tags if needed
5. decide what to do with docker.io, deprecate or publish there too

getzze:
Ok thanks, I'll experiment myself then.

PR #1260, opened 2025-04-10
Deploy docs to gh-pages

getzze:
Also run `CI.yaml` on push tags, maybe that's why Readthedocs was not updating.


getzze:
https://Diaoul.github.io/subliminal is already used as @Diaoul personal website, it cannot be an alternative to Readthedocs...

vivodi:
You could consider transferring the repository to https://github.com/get-subliminal/subliminal. This would allow you to host the documentation directly at https://get-subliminal.github.io, with a cleaner URL structure (no need to place it under a `/subliminal` subpath—just serve it from the site root).

I also noticed that @Diaoul is no longer actively maintaining the project, while @getzze appears to be the current maintainer. Moving the repository under an _organization_ such as `get-subliminal` or `subliminal-tool` would better reflect that the project is no longer individually maintained.

Diaoul:
Yes absolutely, this has been on my to-do for a while!

PR #1362, opened 2026-06-18
Cli: parameterize --name with sed-like substitution

Holek:
Allow `--name` to rewrite each scanned file's name individually instead of passing one static name to every file, removing the need for shell loops. Now `--name 's/pattern/replacement/flags'` applies a sed-style substitution to each file name (back-references \1.., whole-match &, flags g and i).

Files that don't match are scanned as-is. Logic lives in a testable NameResolver in subliminal.utils. Invalid regex/flags raise a click BadParameter. Adds unit and CLI integration tests, docs and changelog.

### DIFF
--- a/docs/user/usage.rst
+++ b/docs/user/usage.rst
@@ -33,31 +33,6 @@ Download English subtitles::
     only. Otherwise you will get banned from the providers for abuse due to too many requests. If subliminal didn't
     find subtitles for an old video, it's unlikely it will find subtitles for that video ever anyway.
 
-Parameterized name
-^^^^^^^^^^^^^^^^^^^
-
-The ``--name/-n`` option lets you override the name used to guess information about a file. In addition to a
-plain static name, it can rewrite each file name individually so you no longer need a shell loop to rename
-poorly-named releases before downloading.
-
-Pass a `sed <https://www.gnu.org/software/sed/manual/sed.html>`_-like substitution
-``s/pattern/replacement/flags`` and it is applied to every file name on its own. Back-references
-(``\1``, ``\2``, …) are available in the replacement, and the ``g`` (replace all occurrences) and ``i``
-(case-insensitive) flags are supported::
-
-    $ subliminal download -l pl \
-        --name 's/.*YP-1R-([0-9]+)x([0-9]+).*/My Little Pony Friendship Is Magic S\1E\2.mkv/' \
-        */YP-1R-*.mkv
-
-Unlike sed, ``&`` is a literal character (not the whole match), so titles containing an ampersand need
-no escaping::
-
-    $ subliminal download -l pl \
-        --name 's/.*_-_([0-9]+)_.*/Panty & Stocking with Garterbelt S01E\1.mkv/' \
-        *Garterbelt_-_*.mkv
-
-Any file whose name does not match is left untouched and scanned as-is.
-
 You can use a configuration file in the `TOML <https://toml.io/en/>`_ format with the ``--config/-c`` option. If no configuration file is
 provided, it looks for a ``subliminal.toml`` file in the default configuration folder for the application. This folder is
 `OS dependent <https://github.com/tox-dev/platformdirs>`_:
@@ -76,6 +51,31 @@ See :ref:`cli` for more details on the available commands and options.
 
         $ python -m subliminal download -h
 
+
+Parameterized name
+^^^^^^^^^^^^^^^^^^^
+
+The ``--name/-n`` option lets you override the file name that is used to guess information about a video.
+In addition to a static alternative name, you can use a
+`sed <https://www.gnu.org/software/sed/manual/sed.html>`_-like substitution pattern
+``s/pattern/replacement/flags`` to override the names of all the files in a folder.
+The pattern needs to be a valid Python regex. Back-references (``\1``, ``\2``, …) are available in the
+replacement, and the ``g`` (replace all occurrences) and ``i`` (case-insensitive) flags are supported::
+
+    $ subliminal download -l pl \
+        --name 's/.*YP-1R-([0-9]+)x([0-9]+).*/My Little Pony Friendship Is Magic S\1E\2.mkv/' \
+        */YP-1R-*.mkv
+
+Unlike sed, ``&`` is a literal character (not the whole match), so titles containing an ampersand need
+no escaping::
+
+    $ subliminal download -l pl \
+        --name 's/.*_-_([0-9]+)_.*/Panty & Stocking with Garterbelt S01E\1.mkv/' \
+        *Garterbelt_-_*.mkv
+
+The substitution is applied to the path of each file (after the optional conversion to an absolute path
+with ``--use-absolute-path``), so the pattern can also match parent directories. Any file whose path does
+not match is left untouched and scanned as-is.
 
 Nautilus/Nemo integration
 -------------------------
--- a/docs/user/usage.rst
+++ b/docs/user/usage.rst
@@ -33,6 +33,32 @@ Download English subtitles::
     only. Otherwise you will get banned from the providers for abuse due to too many requests. If subliminal didn't
     find subtitles for an old video, it's unlikely it will find subtitles for that video ever anyway.
 
+Parameterized name
+^^^^^^^^^^^^^^^^^^^
+
+The ``--name/-n`` option lets you override the name used to guess information about a file. In addition to a
+plain static name, it can rewrite each file name individually so you no longer need a shell loop to rename
+poorly-named releases before downloading.
+
+Use a `sed <https://www.gnu.org/software/sed/manual/sed.html>`_-like substitution
+``s/pattern/replacement/flags`` and it is applied to every file name on its own. Back-references
+(``\1``, ``\2``, …) and the whole-match reference (``&``) are available in the replacement, and the
+``g`` (replace all occurrences) and ``i`` (case-insensitive) flags are supported::
+
+    $ subliminal download -l pl \
+        --name 's/.*YP-1R-([0-9]+)x([0-9]+).*/My Little Pony Friendship Is Magic S\1E\2.mkv/' \
+        */YP-1R-*.mkv
+
+Alternatively, supply a regular expression with ``--name-pattern`` and use ``--name`` as a template whose
+back-references are filled from the match on each file name::
+
+    $ subliminal download -l pl \
+        --name-pattern '.*_-_([0-9]+)_.*' \
+        --name 'Panty & Stocking with Garterbelt S01E\1.mkv' \
+        *Garterbelt_-_*.mkv
+
+Any file whose name does not match is left untouched and scanned as-is.
+
 You can use a configuration file in the `TOML <https://toml.io/en/>`_ format with the ``--config/-c`` option. If no configuration file is
 provided, it looks for a ``subliminal.toml`` file in the default configuration folder for the application. This folder is
 `OS dependent <https://github.com/tox-dev/platformdirs>`_:
--- a/docs/user/usage.rst
+++ b/docs/user/usage.rst
@@ -40,21 +40,20 @@ The ``--name/-n`` option lets you override the name used to guess information ab
 plain static name, it can rewrite each file name individually so you no longer need a shell loop to rename
 poorly-named releases before downloading.
 
-Use a `sed <https://www.gnu.org/software/sed/manual/sed.html>`_-like substitution
+Pass a `sed <https://www.gnu.org/software/sed/manual/sed.html>`_-like substitution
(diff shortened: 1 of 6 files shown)


getzze:
Thanks for the PR, it's a nice extension of the `--name` option. I was wondering if we really need the `--name-pattern`  extra option, allowing sed-like expression in `--name` covers all the use cases, right?

Holek:
Yeah, that makes sense, actually. Since new `--name` is a superset of features, I've opted to remove `--name-pattern`.

docs/user/usage.rst
```diff
@@ -33,6 +33,31 @@ Download English subtitles::
     only. Otherwise you will get banned from the providers for abuse due to too many requests. If subliminal didn't
     find subtitles for an old video, it's unlikely it will find subtitles for that video ever anyway.
 
+Parameterized name
+^^^^^^^^^^^^^^^^^^^
+
+The ``--name/-n`` option lets you override the name used to guess information about a file. In addition to a
+plain static name, it can rewrite each file name individually so you no longer need a shell loop to rename
+poorly-named releases before downloading.
```
getzze:
```suggestion
The ``--name/-n`` option lets you override the file name, that is used to guess information about a video. In addition to a static alternative name, you can use a `sed <https://www.gnu.org/software/sed/manual/sed.html>`_-like substitution pattern ``s/pattern/replacement/flags`` to override the names of every files in a folder. 
The pattern needs to be a valid python regex. Back-references
```

src/subliminal/cli/commands/download_best.py
```diff
@@ -333,7 +332,11 @@
     metavar='NAME',
     help=(
         'Name used instead of the path name for guessing information about the file. '
-        'If used with multiple paths or a directory, `name` is passed to ALL the files.'
+        'If used with multiple paths or a directory, `name` is passed to ALL the files. '
+        'NAME may also be a sed-like substitution `s/pattern/replacement/flags`, applied '
+        r'to each file name individually: back-references (\1, \2, ...) are available in '
+        'the replacement, `&` is a literal character and the `g` (replace all) and `i` '
```
getzze:
```suggestion
        'the replacement, `&` is a literal character (unlike in sed) and the `g` (replace all) and `i` '
```

docs/user/usage.rst
```diff
@@ -33,6 +33,31 @@ Download English subtitles::
     only. Otherwise you will get banned from the providers for abuse due to too many requests. If subliminal didn't
     find subtitles for an old video, it's unlikely it will find subtitles for that video ever anyway.
 
+Parameterized name
```
getzze: Move this section just before the "Nautilus/Nemo integration" section.

src/subliminal/utils.py
```diff
@@ -76,6 +76,151 @@ def safely_guessit(name: str | None, options: dict[str, Any] | None = None) -> d
...
+        i += 1
+    segments.append(''.join(current))
+
+    # Expect exactly pattern, replacement and flags (i.e. three delimiters total)
+    if len(segments) != 3:
+        return None
+    return segments[0], segments[1], segments[2]
+
+
+def _sed_replacement_to_python(replacement: str) -> str:
```
getzze: We don't need this function, "\&" should not be replaced to "&"

tests/test_utils.py
```diff
@@ -364,3 +366,83 @@ def test_clip(value: float, minimum: float | None, maximum: float | None, expect
...
+    assert NameResolver('s/a/X/g')('aaa.mkv') == 'XXX.mkv'
+    # i is case-insensitive
+    assert NameResolver('s/a/X/gi')('AaA.mkv') == 'XXX.mkv'
+
+
+def test_name_resolver_sed_ampersand_is_literal() -> None:
+    # unlike sed, & is a literal character (not the whole match)
+    assert NameResolver('s/[0-9]+/[&]/')('ep12.mkv') == 'ep[&].mkv'
+    # an escaped \& also yields a literal &
+    assert NameResolver(r's/[0-9]+/\&/')('ep12.mkv') == 'ep&.mkv'
```
getzze: this test should be modified

tests/cli/test_download.py
```diff
@@ -538,3 +538,101 @@ def test_cli_download_with_generated_config(cli_runner: CliRunner) -> None:
         # collect files recursively
         files = [os.fspath(p.relative_to(td)) for p in Path(td).rglob('*')]
         assert subtitle_filename in files
+
```
getzze, line 541: add a test for a static name. ideally the static name should start with "s" to see if the "static" mode was well identified.

src/subliminal/utils.py
```diff
@@ -76,6 +76,151 @@ def safely_guessit(name: str | None, options: dict[str, Any] | None = None) -> d
...
+            i += 1
+            continue
+        current.append(c)
+        i += 1
+    segments.append(''.join(current))
+
+    # Expect exactly pattern, replacement and flags (i.e. three delimiters total)
+    if len(segments) != 3:
+        return None
+    return segments[0], segments[1], segments[2]
```
getzze, line 134:
```suggestion
def split_esc(string: str, delimiter: str) -> Iterator[str]:
    """Split the string by the non-escaped delimiter.

    Return an iterator of the string parts.
    """
    dln = len(delimiter)
    ln = len(string)
    i = 0
    j = 0
    while j < ln:
        # Delimiter found
        if string[j:j + dln] == delimiter:
            yield string[i:j]
            j += dln
            i = j
        # Escaped character
        elif string[j] == '\\':
            if j + 1 >= ln:
                yield string[i:j]
                return
            j += 2
        # Other character
        else:
            j += 1
    yield string[i:j]


def parse_sed_expression(expr: str) -> tuple[str, str, str] | None:
    """Parse a sed-like substitution ``s<delim>pattern<delim>replacement<delim>flags``.

    The delimiter is the character right after the leading ``s`` and may be any
    non-alphanumeric, non-backslash, non-whitespace character (commonly ``/``).
    A delimiter escaped with a backslash inside the pattern or the replacement is
    treated as a literal delimiter character.
    :param str expr: the expression to parse.
    :return: ``(pattern, replacement, flags)`` or ``None`` if `expr` is not a
        well-formed sed substitution.
    :rtype: tuple of (str, str, str) or None
    """
    if len(expr) < 2 or expr[0] != 's':
        return None
    delim = expr[1]
    if delim.isalnum() or delim == '\\' or delim.isspace():
        return None

    segments = list(split_esc(expr[2:], delim)
    # Expect exactly pattern, replacement and flags (i.e. three delimiters total)
    if len(segments) != 3:
        return None
    return segments[0], segments[1], segments[2]

```

getzze: Let's try this implementation, that is a bit more readable to me :)

src/subliminal/utils.py
```diff
@@ -76,6 +76,151 @@ def safely_guessit(name: str | None, options: dict[str, Any] | None = None) -> d
...
+      and ``i`` (case-insensitive match).
+
+    In the sed mode, a file whose base name does not match is left untouched
+    (``None`` is returned, and the original path is used as before).
+
+    :param str name: the ``--name`` value, may be ``None``.
+    :raises ValueError: if the sed expression contains an invalid regular
+        expression or an unsupported flag.
+    """
+
```
getzze, line 158: add attributes types here

src/subliminal/utils.py
```diff
@@ -76,6 +76,151 @@ def safely_guessit(name: str | None, options: dict[str, Any] | None = None) -> d
...
+        self._replacement = _sed_replacement_to_python(replacement)
+        self._count = 0 if 'g' in flags else 1
+        self.mode = 'sed'
+
+    def __call__(self, filepath: str | os.PathLike[str]) -> str | None:
+        """Return the name to use for `filepath`, or ``None`` to use the path as-is."""
+        if self.mode == 'static' or self._regex is None:
+            return self.name
+
+        base = os.path.basename(os.fspath(filepath))
```
getzze: This is a bit restrictive, see the solution of making the "name" argument of "scan_video_path" a of `NameResolver` type.

src/subliminal/cli/commands/download_best.py
```diff
@@ -467,15 +476,19 @@ def download(
             # scan videos
             video_candidates: list[Video] = []
             for filepath in collected_filepaths:
+                # Resolve the name to use for this specific file
+                file_name = name_resolver(filepath)
                 # Try scanning the video at path
-                video = scan_video_path(filepath, absolute_path=absolute_path, name=name, verbose=verbose, debug=debug)
+                video = scan_video_path(
+                    filepath, absolute_path=absolute_path, name=file_name, verbose=verbose, debug=debug
+                )
```
getzze, line 482: Let's pass `name_resolver` to the "name" argument to deal with "absolute_path=True/False" in the "scann_video_path" function

getzze (changes requested):
It looks good, I have some comments, the most important is to which string the substitution is applied. I think it should respect the "use_absolute_path" option.

getzze:
I'm thinking that probably we should allow multiple use of the `--name` option, to allow several substitution at the same time.

[Edit] it's better to do it in a separate PR

docs/user/usage.rst
Holek: Applied in 2e930b0, with small grammar tweaks ("to override the names of all the files in a folder", "valid Python regex").

src/subliminal/cli/commands/download_best.py
Holek: Applied in 2e930b0.

docs/user/usage.rst
Holek: Moved in 2e930b0.

src/subliminal/utils.py
Holek: Removed in 2e930b0 — the replacement is now passed verbatim to `re.sub`, which leaves `\&` alone, so it stays as a literal `\&` in the output.

tests/test_utils.py
Holek: Updated in 2e930b0 to assert that `\&` is kept verbatim.

tests/cli/test_download.py
Holek: Added `test_cli_download_name_static` in 2e930b0. The static name is `s.w.a.t.2017.s01e01.720p.hdtv.x264-killers.mkv`: it starts with `s` and its second character is non-alphanumeric, so it actually goes through the sed parsing attempt (splitting into 9 segments) before falling back to static mode.

src/subliminal/utils.py
Holek: Adopted in 2e930b0 (with one fix: a missing closing parenthesis in `list(split_esc(expr[2:], delim))`). One semantic difference from the previous parser: an escaped delimiter now keeps its backslash in the segments, e.g. `s/a\/b/c/` yields the pattern `a\/b` — the regex engine treats it as a literal `/` anyway. Tests updated accordingly, and `split_esc` got its own unit tests.

src/subliminal/utils.py
Holek: Added in 2e930b0.

src/subliminal/utils.py
Holek: Changed in 2e930b0: the substitution is now applied to the whole path, and `scan_video_path` accepts a `NameResolver` as its `name` argument.

src/subliminal/cli/commands/download_best.py
Holek: Done in 2e930b0 — `scan_video_path` now takes the resolver and resolves the name after the conversion to absolute path, so the substitution respects `--use-absolute-path` (including the `fallback` mode, where the retry applies the pattern to the absolute path).

Holek:
Thanks for the thorough review! All comments are addressed in 2e930b0. The main change: the substitution is now applied to the whole file path, resolved inside `scan_video_path` after the absolute-path conversion, so it respects `--use-absolute-path` in all three modes. Also switched to the `split_esc`-based parser, dropped the `\&` unescaping, added the static-name test and moved/reworded the docs section.

src/subliminal/utils.py
getzze: Mmm, then that's incorrect, an escaped delimiter should be unescaped (remove the backslash).
In the "elif string[j] == '\\':" branch, we can test if the next characters are a delimiter, if yes, do.not yield the '\\'.

getzze:
Thanks for the changes, I think there is just a small change needed in `split_esc` (and addind the relevant tests).

src/subliminal/utils.py
getzze: I had a better look at the code, the best would be to replace the escaped delimiters by delimiters in the output of the `split_esc` iterator.

getzze (approved):
I made the last modifications, it's good to go. Thanks again for your contribution.

Issue #1275, opened 2025-04-30
2.3.0: pytest fails with missing win32_setctime on unix oses


nunotexbsd:
FreeBSD / Python 3.11

I've ported `knowit` and `trakit` (not committed yet) and subliminal is using following test dependencies:

```
TEST_DEPENDS=   ${PYTHON_PKGNAMEPREFIX}defusedxml>0:devel/py-defusedxml@${PY_FLAVOR} \
                ${PYTHON_PKGNAMEPREFIX}knowit>0:multimedia/py-knowit@${PY_FLAVOR} \
                ${PYTHON_PKGNAMEPREFIX}sympy>0:math/py-sympy@${PY_FLAVOR} \
                ${PYTHON_PKGNAMEPREFIX}tomlkit>0:textproc/py-tomlkit@${PY_FLAVOR} \
                ${PYTHON_PKGNAMEPREFIX}vcrpy>0:devel/py-vcrpy@${PY_FLAVOR}

```
 
I'm in doubt if I need to port `win32_setctime` too as it seems to be windows related?
I there a quick workaround to avoid it?

Thanks

```
platform freebsd15 -- Python 3.11.12, pytest-8.1.1, pluggy-1.5.0 -- /usr/local/bin/python3.11
cachedir: .pytest_cache
rootdir: /wrkdirs/usr/ports/multimedia/py-subliminal/work-py311/subliminal-2.3.0
configfile: pyproject.toml
collected 590 items / 1 error / 7 deselected / 583 selected

====================================================================================================== ERRORS ======================================================================================================
_______________________________________________________________________________________ ERROR collecting tests/test_video.py _______________________________________________________________________________________
ImportError while importing test module '/wrkdirs/usr/ports/multimedia/py-subliminal/work-py311/subliminal-2.3.0/tests/test_video.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/local/lib/python3.11/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/test_video.py:10: in <module>
    from win32_setctime import SUPPORTED, setctime
E   ModuleNotFoundError: No module named 'win32_setctime'
[...]

Issue #1157, opened 2024-07-21
2.2.1 testunit error


nunotexbsd:
1 error on testunit that I can't figure out if it is missing some dependency.

Any clues?
Thanks

```
platform freebsd15 -- Python 3.11.9, pytest-8.1.1, pluggy-1.5.0 -- /usr/local/bin/python3.11
cachedir: .pytest_cache
rootdir: /wrkdirs/usr/ports/multimedia/py-subliminal/work-py311/subliminal-2.2.1
configfile: pyproject.toml
testpaths: tests
collected 367 items / 1 error

====================================================================================================== ERRORS ======================================================================================================
_______________________________________________________________________________________ ERROR collecting tests/test_score.py _______________________________________________________________________________________
ImportError while importing test module '/wrkdirs/usr/ports/multimedia/py-subliminal/work-py311/subliminal-2.2.1/tests/test_score.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/local/lib/python3.11/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/test_score.py:7: in <module>
    from subliminal.score import compute_score, episode_scores, movie_scores, solve_episode_equations, solve_movie_equations
E   ImportError: cannot import name 'solve_episode_equations' from 'subliminal.score' (/wrkdirs/usr/ports/multimedia/py-subliminal/work-py311/subliminal-2.2.1/subliminal/score.py)
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
================================================================================================= 1 error in 0.16s =================================================================================================
*** Error code 2
```

getzze:
It looks like the `sympy` test dependency is missing.
You need to install subliminal with the `test` optional dependencies:
`python -m pip install -e '.[test,dev]'`

`CONTRIBUTING.md` is incorrect, I will update it. Thanks!

nunotexbsd:
@getzze 

That was a missing dep that I missed from pyproject.py. It works now.

[...]

PR #1370, opened 2026-07-22
Add CI for FreeBSD

getzze:
@nunotexbsd, this should be enough to check for bugs with FreeBSD, no?


getzze:
For back-reference, bugs specific to FreeBSD: #1369 #1275 #1157

getzze:
It takes 5min to run, so I would prefer to run it only on pushes, not pull_request.

PR #1376, opened 2026-08-14
fix(opensubtitlescom): prevent pagination errors from discarding collected results

### LINKED ISSUE #1373: opensubtitlescom: an error on one page discards the results of all earlier pages
## What happens

`_search` is a generator that fetches one page per iteration and yields its rows. The pagination loop has no error handling.

`api_get` raises `OpenSubtitlesComError` for any HTTP status other than 200.
When a page request fails, the exception propagates out of the generator into `query`.

`query` accumulates subtitles in a local list while it iterates the generator. The caller gets an exception instead of the subtitles from the pages that succeeded.

In practice the server returns HTTP 400 at around page 21 (probably a server-side page limit?)
`max_result_pages` defaults to 0 (unlimited), so it does not cap the loop before that point.

## How to reproduce

The stubbed `api_get` returns two pages and then fails, like the live API does:

```python
from subliminal.providers.opensubtitlescom import OpenSubtitlesComError, OpenSubtitlesComProvider

provider = OpenSubtitlesComProvider.__new__(OpenSubtitlesComProvider)
provider.max_result_pages = 0

def fake_api_get(path, params, **kwargs):
    page = params.get('page', 1)
    if page >= 3:
        raise OpenSubtitlesComError('Bad Request')
    return {'data': [{'id': f's{page}-{i}'} for i in range(3)], 'total_pages': 100}

provider.api_get = fake_api_get

list(provider._search(languages='en'))
```

Result: `OpenSubtitlesComError: Bad Request`.
The six results from pages 1 and 2 are lost.

## Expected behaviour

Pagination stops at the failing page and the caller keeps the results collected so far.

## Suggested fix

Catch `ProviderError` inside the pagination loop and break:

```python
try:
    response = self.api_get('subtitles', ext_params)
except ProviderError:
    logger.warning('Pagination stopped at page %d', page)
    break
```

## Context

It seems to be a regression. PR [#1336](https://github.com/Diaoul/subliminal/pull/1336) rewrote `_search` from recursion to a generator in 2.6.0. The rewrite did not add error handling.

wujekbogdan:
Fixes #1373 - a bug that caused opensubtitles.com search pagination errors to discard all results collected on the pages that succeeded. The fix prevents errors from being re-raised if some results have already been collected by previous iterations.

### DIFF
--- a/changelog.d/1373.provider.rst
+++ b/changelog.d/1373.provider.rst
@@ -1,0 +1,1 @@
+Stop the opensubtitlescom pagination when a page fails and keep the subtitles already found
--- a/src/subliminal/providers/opensubtitlescom.py
+++ b/src/subliminal/providers/opensubtitlescom.py
@@ -13,12 +13,14 @@ from typing import TYPE_CHECKING, Any, ClassVar, cast
 from babelfish import Language, language_converters  # type: ignore[import-untyped]
 from dogpile.cache.api import NO_VALUE
 from requests import Response, Session
+from requests.exceptions import RequestException
 
 from subliminal import __short_version__
 from subliminal.cache import region
 from subliminal.exceptions import (
     AuthenticationError,
     ConfigurationError,
+    DiscardingError,
     DownloadLimitExceeded,
     NotInitializedProviderError,
     ProviderError,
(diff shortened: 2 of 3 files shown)


getzze:
Thanks for the PR, can you rebase on main, I pushed fixes for the failing CI.

Then it can be merged.

wujekbogdan:
Done.

PR #1378, opened 2026-08-15
feat(core): report the providers that failed during a search

### LINKED ISSUE #1374: ProviderPool: a provider that fails with an unexpected error is not reported to the caller
This is a behaviour change request, not a bug report.

## Context

I am working on a subtitle plugin which uses subliminal as a library rather than through the CLI. It searches a set of providers and shows the results to a user.

When a provider breaks, I want to tell the user something useful, like "opensubtitlescom failed", instead of showing an empty list. Right now I can't, because the pool gives me no way to distinguish errors from valid but empty results.

## What happens

`list_subtitles_provider` splits its exception handling in two. A `DiscardingError` returns `None`, and the pool adds the provider to `discarded_providers`. Any other exception returns an empty list, and the provider is treated as if it had searched successfully and found nothing.

The split is deliberate - the code says so: `return [] so the provider is not discarded with unknown error` so an empty list from a provider that crashed looks exactly like an empty list from a provider that had no matches.

The pool is already inconsistent about this

`download_subtitle` in the same class discards on *any* exception:

```python
except Exception as e:
    handle_exception(e, f'Discarding provider {subtitle.provider_name}')
    self.discarded_providers.add(subtitle.provider_name)
```

So a provider that fails while downloading is discarded, and the same provider failing while listing is not.
The CLI warning text says `Some providers have been discarded due to unexpected errors`, but today only `DiscardingError`, which is an *expected* error, ever triggers it.

## Impact on the CLI

It's not a breaking change for CLI users. `download_best` uses `AsyncProviderPool`, whose `list_subtitles` ignores `discarded_providers`, so every provider is still queried for every video. No new exceptions, and no change to output or exit codes.

## Request

**Option 1: discard on every exception.**

It's simple, and it matches what `download_subtitle` already does:

```python
except Exception as e:
    handle_exception(e, f'Provider {provider}')
    return None
```

**Option 2 - report failures separately.**

Keep `discarded_providers` as it is, and add a set that records which providers failed without changing what the pool skips.

wujekbogdan:
Resolves #1374

`ProviderPool` reported a provider in `discarded_providers` only when it raised a `DiscardingError`. Any other exception was logged, and the provider returned an empty list. A caller could not distinguish a provider that crashed from a provider that found nothing.

Added `ProviderPool.failed_providers`, a set that records every provider that had an error. A discarded provider always had an error, so `discarded_providers` is a subset of it.

The search behaviour and the download behaviour are untouched, so the change is backward compatible for library users and for the CLI.

The `download` command warned about discarded providers already. Now it also names every provider that had an error, and tells the user how these errors affect the download.


changelog.d/1374.change.rst
```diff
@@ -0,0 +1,2 @@
+Added `ProviderPool.failed_providers` - a set of providers that had an error during a search.
+Changed the user-facing error messages of the `download` command to explicitly distinguish discarded providers, which are not used for the rest of the download, from those that errored out but were still kept in the pool.
```
wujekbogdan: I'm not sure changelogs should be this verbose. I tried to keep it short, but at the same time I wanted to make it very clear exactly what changed.

src/subliminal/core.py
```diff
@@ -117,6 +121,14 @@ def __delitem__(self, name: str) -> None:
     def __iter__(self) -> Iterator[str]:
         return iter(self.initialized_providers)
 
+    def _discard_provider(self, name: str) -> None:
+        """Stop using a provider, and record its error.
+
+        A discarded provider is always a failed provider - that's why one function writes the two sets.
+        """
+        self.failed_providers.add(name)
+        self.discarded_providers.add(name)
```
wujekbogdan, line 130: I know it feels like an unnecessary abstraction, since these are just two lines that don't seem to deserve a function, but my motivation was that these two calls should always happen together: `failed_providers` is a superset of `discarded_providers` (a discarded provider has always failed, but not vice versa).

tests/cli/test_format.py
```diff
@@ -0,0 +1,48 @@
...
+        (
+            {'gestdown', 'podnapisi', 'tvsubtitles'},
+            {'gestdown'},
+            3,
+            'These providers had an error and were skipped for the rest of the download: gestdown. '
+            'These providers had an error on some videos: podnapisi, tvsubtitles. '
+            'Some subtitles can be missing.',
+        ),
+    ],
+)
```
wujekbogdan, line 46: It took me a while to figure out a proper format for user-facing messages. On one hand I wanted them to be precise, on the other hand I didn't want to make them too verbose and expose internal implementation details that aren't relevant, or are simply too low-level to be useful for end users.

Basically, what I wanted to convey was that a discarded provider is discarded for good, while a failed one only failed for a particular video and can still be used for subsequent downloads.

I didn't use the discarded/failed terminology on purpose, because it means nothing to end users who aren't familiar with the codebase and its internal terminology.

I hope I found the right balance.

src/subliminal/core.py
```diff
@@ -150,9 +162,10 @@ def list_subtitles_provider(self, provider: str, video: Video, languages: Set[La
             handle_exception(e, f'Provider {provider}')
             # return None to discard this provider with a known error
             return None
-        except Exception as e:  # noqa: BLE001  # pragma: no cover
+        except Exception as e:  # noqa: BLE001
             handle_exception(e, f'Provider {provider}')
-            # return [] so the provider is not discarded with unknown error
+            # an unknown error can come from one video only, thus keep the provider for the next ones
+            self.failed_providers.add(provider)
```
getzze, line 168: the `list_subtitles_provider` method is used by the multi-threaded class `AsyncProviderPool`. Therefore is needs to have no side-effect...
Maybe a solution would be that `list_subtitles_provider` return a list of subtitles on success, otherwise the exception. And the exception is processed downstream.

changelog.d/1374.change.rst
getzze: I think it's fine, it's better to explain the changes well

src/subliminal/cli/commands/_format.py
```diff
@@ -54,6 +57,22 @@ def convert(self, value: str, param: click.Parameter | None, ctx: click.Context
...
+def format_provider_errors(failed_providers: Set[str], discarded_providers: Set[str], *, video_count: int) -> str:
+    """Tell the user which providers had an error, and how the error affected the end results.
+
+    With one video both kinds of error have the same result, so one sentence covers them.
+    """
+    if video_count == 1:
+        return f'These providers had an error and found no subtitles: {", ".join(sorted(failed_providers))}.'
+
+    groups = (
+        ('These providers had an error and were skipped for the rest of the download', sorted(discarded_providers)),
```
getzze, line 69:
```suggestion
        ('These providers had an error and were discarded for the rest of the download', sorted(discarded_providers)),
```

getzze (commented):
Thanks, error handling in the pool is a bit weak indeed.
As you mentioned, `download_subtitle` should also filter discarded/failed providers.

src/subliminal/cli/commands/_format.py
wujekbogdan: See my comment here: https://github.com/Diaoul/subliminal/pull/1378#discussion_r3789435874 

I avoided the `discarded` terminology on purpose. It's a valid term for programming, but IMO it's not a great user-facing term, since it doesn't really explain what happened. `skipped` explains the outcome, `discarded` just explains the internal "action" we took.

getzze: ok, you're right

getzze:
This is a deeper problem than I thought, let me first make add a PR to introduce a `ProviderResult` class and then add this feature on top.

[EDIT] #1381

getzze:
#1381 was merged, you can rebase and update this PR, thanks!

src/subliminal/core.py
```diff
@@ -116,7 +119,17 @@ def __init__(
...
+
+    @property
+    def discarded_providers(self) -> frozenset[str]:
+        """Providers that had an error that makes them unusable - they're discarded once the error is reported."""
+        return frozenset(r.provider for r in self.results if r.outage())
+
+    @property
+    def failed_providers(self) -> frozenset[str]:
+        """Providers that had an error, discarded or not. A superset of :attr:`discarded_providers`."""
+        return frozenset(r.provider for r in self.results if r.failure() or r.outage())
```
wujekbogdan, line 135: Thanks to [results](https://github.com/Diaoul/subliminal/pull/1378/changes#diff-0ffa83696770e4c227365068beeaff6016bd1aaf90a2250147063996b9b74f3eR112), I could make these two derived from it. IMO that's cleaner than a mutable `set()`, and as a bonus they're read-only, which makes them harder to misuse for consumers of this class.

src/subliminal/core.py
```diff
@@ -80,6 +81,11 @@ def outage(self) -> bool:
         """Check if the query triggered an error that requires discarding the provider."""
         return self.status == ProviderResultStatus.OUTAGE
 
+    @classmethod
+    def from_download(cls, subtitle: Subtitle, status: ProviderResultStatus) -> ProviderResult:
+        """Result of the download of a subtitle."""
+        return cls(subtitle.provider_name, None, {subtitle.language}, [subtitle], status=status)
+
```
wujekbogdan, line 88: Initially I made it a static method of `ProviderPool`, but since it didn't read any class properties, I moved it here. It changes the `ProviderResult` contract though, so let me know if that works for you.

wujekbogdan:
@getzze 

> can rebase and update this PR, thanks!

Done. Could you have a look?

src/subliminal/cli/commands/_format.py
```diff
@@ -54,6 +57,22 @@ def convert(self, value: str, param: click.Parameter | None, ctx: click.Context
         return timedelta(**{k: int(v) for k, v in match.groupdict(0).items()})
 
 
+def format_provider_errors(failed_providers: Set[str], discarded_providers: Set[str], *, video_count: int) -> str:
```
getzze, line 60: This could be a method in `ProviderPool`, it will be easier to find it later.

getzze (commented):
I'm wondering if it's a good idea to keep the list of all results in memory. Do we really need it?
I would prefer to keep it simple and have two normal attributes `discarded_providers` and `failed_providers`. Or do you need them?

wujekbogdan:
@getzze 

> I'm wondering if it's a good idea to keep the list of all results in memory. Do we really need it?
I would prefer to keep it simple and have two normal attributes discarded_providers and failed_providers. Or do you need them?

That's more of a personal preference. I'm into functional programming and immutable data structures, so I always lean towards immutability - which in turn implies derived state.

But you're the boss here - I'll follow your recommendation :)

src/subliminal/cli/commands/_format.py
wujekbogdan: The reason I moved it to a standalone function is that `ProviderPool` is the library's domain object. Nothing in `core.py` "speaks" to the end user currently - it only logs.

I understand that your motivation is discoverability, but its only consumer today is `download_best.py`, and the other formatting helpers already live there.

If I were to refactor it, I'd rather have a class that takes a `ProviderPool` and turns it into human-readable text (like a presenter).

---

Anyway, I won't die on that hill - you're the boss, it's your call.

getzze: Good point! Let's leave it like that then

getzze:
I thought about it and for now let's keep the `discarded_providers` and `failed_providers` attributes and not keep the results. Like this we also don't need the `from_download` classmethod that changed the `ProviderResult` too much.

getzze (approved):
This is ready to be merged, if you don't need to add more commits

wujekbogdan:
Thanks for your assistance. Let's merge it!

PR #1381, opened 2026-09-02
Add a ProviderResult class to keep track of queries to providers

getzze:
First step to solve #1374 

Keep track of the query and result (list of subtitles or exception) of the `ProviderPool.list_subtitles_provider` method, to be used downstream to keep track of discarded or failed providers.

### DIFF
--- a/changelog.d/1381.change.rst
+++ b/changelog.d/1381.change.rst
@@ -1,0 +1,1 @@
+add ProviderResult class to keep track of queries to providers
--- a/src/subliminal/core.py
+++ b/src/subliminal/core.py
@@ -9,6 +9,7 @@ import os
 from collections import defaultdict
 from concurrent.futures import ThreadPoolExecutor
 from dataclasses import dataclass, field
+from enum import Enum, auto
 from typing import TYPE_CHECKING, Any
 
 from babelfish import Language  # type: ignore[import-untyped]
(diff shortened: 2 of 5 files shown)


wujekbogdan:
Thank you - that's much better!

I'm just wondering if it wouldn't be even better to provide an enum for the results. `list_subtitles_provider` already catches everything and now carries the exception as a value. So these are no longer exceptions in the literal sense - they are values of an exception type.

Also, in the current format we can't easily tell what the reason was. A `ProviderQueryError` covers two different cases: wrong video type and unsupported language, and the only way to find out what truly happened is to parse the message. 

For the other cases the reader has to know the class hierarchy: `DiscardingError` means discard, `None` means success, anything else is a failure.

How about we map them into an enum that carries the meaning explicitly, e.g:

```python
class ProviderStatus(Enum):
    OK = auto()
    UNSUPPORTED_VIDEO = auto()
    UNSUPPORTED_LANGUAGES = auto()
    DISCARDED = auto()
    FAILED = auto()
```

`ProviderResult` will get a `status` field, and `exception` would stay as it is for the details of `DISCARDED` and `FAILED ` (or we could drop it, idk)

getzze:
That's a good idea, thanks!
I will remove the `exception` attribute, its message was printed anyway, so we don't really need it anymore.

getzze:
Done! I also fixed the fact that any error when downloading subtitle was discarding.

I struggled for the status name associated with an error that leads to discard the provider. I ended up using "Outage", a term used in safety and security systems, to differentiate it from a "Failure".

